PR #1, opened 2018-06-21
Commit initial copy of members list

cblecker:


fejta:
No team info?

fejta:
Also I think we should leave of billing email

cblecker:
Removed billing e-mail.

Yes, I left off the teams on purpose. I'm hoping to be able to split up the teams based on sig.. but also something I noticed is that org owners are always listed as a "maintainer" of all teams they're on.

cblecker:
If we need to change the format at a later date, then we can.. but I want to get an initial public dump for https://github.com/cncf/devstats/issues/124

/approve

PR #2, opened 2018-07-06
Update meta org

cblecker:
This updates the meta org as follows:
- Pull a copy of the OWNERS_ALIASES from k/community
- Updates membership as of 7/16
- Adds a folder of team members for sig-contributor-experience. Example of what it would look like to have sig-responsible teams.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,6 +1,152 @@
-# See the OWNERS docs: https://git.k8s.io/community/contributors/devel/owners.md
-
 aliases:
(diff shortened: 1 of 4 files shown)


fejta:
/lgtm
/hold

I think we can do something like having a `teamsFile:` field instead of `teams`. So it would be something like:

```
orgs:
  kubernetes:
    members: ["cblecker", "fejta"]
    teamsFiles:
    - prefix: sig-contributor-experience # Ensure all names start with this
       path: ./sig-contributor-experience/org.yaml
```

PR #8, opened 2018-08-11
Script to apply declared config

fejta:
Plus other initial infrastructure

/assign @spiffxp @cblecker 

Primarily supporting the following:
```shell
bazel test //config:all  # Test all the various config/kubernetes-foo orgs
bazel run //admin:update # Run peribolos on all the files
```

Docs at https://github.com/fejta/org/blob/infra/admin/README.md


fejta:
/approve
🙄

cblecker:
/hold

Because we don't have CI in this repo yet, I want to review this closely before merge. Thanks for doing this!

fejta:
Sure I can also move the vendor (or any of the other commits) into their own PRs if that helps. 

Also I can start setting up ci after this goes in. 


We are adding all the verify-boilerplate etc stuff

fejta:
ping

fejta:
FYI moved the first two commits into https://github.com/kubernetes/org/pull/27

fejta:
okay this is now ready to go and < 500 LOC

PR #28, opened 2018-08-21
Add some bazel and hack rules

fejta:
Once these are checked in I can add presubmit jobs for this repo

/assign @calebamiles @spiffxp @cblecker


cblecker:
/lgtm
/hold

🚀

fejta:
Some concerns about excess nosiness around the `set -o xtrace` and we may want to remove these later but can be a follow up PR

fejta:
/hold cancel

Thanks for the review!

PR #41, opened 2018-08-23
Allow hack/*-gofmt.sh to work without installing gosdk

fejta:
Also ensures the correct version of gofmt is used

/hold
Want this PR to pass presubmits after https://github.com/kubernetes/test-infra/pull/9133 merges


fejta:
Now let's pass tess

Issue #16, opened 2018-08-20
REQUEST: membership for kubernetes-incubator/reference-docs/OWNERS


spiffxp:
### GitHub Usernames
@Rajakavitha1
@bradamant3
@bradtopol
@chenopis
@kbarnard10
@mistyhacks
@ryanmcginnis
@stewart-yu
@tengqm
@xiangpengzhao
@zacharysarah
@zhangxiaoyu-zidif
@zparnold

### Organization you are requesting membership in
kubernetes-incubator

### Requirements
I have verified all of the above users are members of kubernetes

### Sponsors
n/a

### List of contributions to the Kubernetes project
n/a

ref: https://github.com/kubernetes-incubator/reference-docs/pull/56

cblecker:
LGTM

spiffxp:
/close
invites sent

PR #98, opened 2018-09-19
Add additional users to kubernetes-incubator

cblecker:
There were some users in kubernetes/org#16 that didn't end up getting PR'd and didn't accept their invite right away. This adds them to the config so they don't get deleted as soon as they accept their invite.


fejta:
It will also delete invitations I believe it the user isn't in the config

PR #114, opened 2018-09-24
Require org members to be sorted

fejta:
/assign @cblecker @spiffxp @nikhiljindal 

a) Do we want to enforce this (we already enforce the absence of duplicates)
b) Do we want to enforce case sensitive sorting?

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


spiffxp:
/lgtm
I am fine with this

cblecker:
Enforcing sorting: 👍
Enforcing case sensitive sorting: 👎

Main thought behind that, is the dump from peribolos doesn't dump in case sensitive sorting (I'm guessing that's simply the way GitHub returns results). We also "normalize" case in a lot of places, so I'd like to ignore case wherever possible.

fejta:
PTAL, now require sorting to be case insensitive

cblecker:
/lgtm
/approve
/hold

🎉 🚀 🚢

nikhita:
/unassign @nikhiljindal 

lgtm :+1:

PR #165, opened 2018-10-12
Update the repo-create template with more questions

spiffxp:
These are questions I find myself guessing the answer to or
having to ask as followup requests to the initial create

I'm also adding "staging" as an option since we tend to
create a bunch of those and they have slightly different
requirements from new or migrated repos

/cc @cblecker @calebamiles
Since you've also handled repo creation in the past

### DIFF
--- a/.github/ISSUE_TEMPLATE/repo-create.md
+++ b/.github/ISSUE_TEMPLATE/repo-create.md
@@ -4,7 +4,7 @@ about: Create or migrate a repository into a Kubernetes Org
 
 ---
 
-### New Repo, or migrate existing
+### New Repo, Staging Repo, or migrate existing
 e.g. new repository
 
 ### Requested name for new repository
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/repo-create.md
```diff
@@ -13,15 +13,36 @@ e.g. cloud-provider-foo
...
+Please prove you have followed the appropriate approval process for this new
+repo by including links to the relevant approvals (meeting minutes, e-mail
+thread, etc.)
+
+Authoritative requirements are here: https://git.k8s.io/community/github-management/kubernetes-repositories.md
+
+tl;dr (but really you should read the linked doc, this may be stale)
+- If this is a core repository, then sig-architecture must approve
+- If this is a SIG repository, then this must follow the procedue spelled out
+in your SIG's charterthe SIG Technical Leads must approve
```
cblecker: this doesn't look right

spiffxp: good catch, that was intended to be an e.g. which I have decided to drop in favor of "just check the charter"

spiffxp:
@cblecker updated, PTAL

.github/ISSUE_TEMPLATE/repo-create.md
```diff
@@ -13,15 +13,36 @@ e.g. cloud-provider-foo
...
+e.g. chris, dianne
+
+### If a new repo, who should be listed as approvers in OWNERS
+e.g. alice, bob
+
+### If a new repo, who should be listed in SECURITY_CONTACTS
+e.g. alice, bob
+
+### What should the repo description be
+by default we will set it to "created due to (link to this issue)"
```
cblecker: I don't like using the link as a default description, but not strongly enough to actually hold this up.

Issue #184, opened 2018-10-23
REQUEST: New kubernetes-sigs membership for liyinan926


liyinan926:
### GitHub Username
@liyinan926 

### Organization you are requesting membership in
kubernetes-sigs

### Requirements
- [x] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [x] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [x] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [x] I am actively contributing to 1 or more Kubernetes subprojects
- [x] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [x] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- Already a member of kubernetes.

### List of contributions to the Kubernetes project
- PRs reviewed / authored
https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Aliyinan926+archived%3Afalse+is%3Aclosed+org%3Akubernetes+
helm/charts#6239
kubernetes-sigs/application#55
- Issues responded to
kubernetes-sigs/application#57
kubernetes-sigs/kubebuilder#453
- SIG projects I am involved with
kubernetes-sigs/application

/area github-membership

Issue #179, opened 2018-10-22
Request for owner/admin access to cri-o under kubernetes-sig


rhatdan:
GitHub Username
@rhatdan
@giuseppe

Organization you are requesting membership in
kubernetes-sigs

Requirements
  I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
  I have enabled 2FA on my GitHub account (https://github.com/settings/security)
  I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
  I am actively contributing to 1 or more Kubernetes subprojects
  I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
  I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application
Sponsors
@runcom 
@mrunalp 

SIG projects I am involved with
CRI-O

mrunalp:
@spiffxp ptal. @rhatdan and @giuseppe are CRI-O maintainers. Thanks!

justaugustus:
+1 as a sponsor! Thanks for your work on CRI-O! :)

I've opened #185 to add Dan and Giuseppe to `kubernetes-sigs`. 
Once that merges, we can get them the appropriate permissions on https://github.com/kubernetes-sigs/cri-o

justaugustus:
@rhatdan @giuseppe -- I'd also like to see a non Red Hat sponsor, for the sake of company diversity. Can you track one down?

vrothberg:
Consider me (SUSE employee) as a sponsor of @giuseppe and @rhatdan.

justaugustus:
@vrothberg thanks for volunteering! Unfortunately, the additional sponsor would have to be an existing member of a Kubernetes org.

sameo:
@justaugustus Please add me as a sponsor there. I'm part of the kubernetes-incubator org.

justaugustus:
LGTM! Thanks @sameo :)

cblecker:
Noting that this is related to the migration of cri-o: https://github.com/kubernetes/org/issues/60

(Dug into the history of this one)

rhatdan:
Well all of us were members of the kubernetes-incubator and need to move over to kubernetes-sigs.

cblecker:
Closing this as invites have been sent out :)

/close

cblecker:
[...]

PR #185, opened 2018-10-23
Add new members (181023)

justaugustus:
Fixes: #184 
Related: #179


justaugustus:
@cblecker -- another non-RH sponsor has been added to #179, so this can be approved now.
/hold cancel

PR #191, opened 2018-10-24
Add --merge-teams and --ignore-teams flags to the merge binary

fejta:
/assign @cblecker @spiffxp 

This will allow us to delegate team membership to sigs.

Try out with `bazel run //cmd/merge -- --org-part=kubernetes=$PWD/config/kubernetes/org.yaml --merge-teams`

When we want to turn this on switch `--ignore-teams` to `--merge-teams` in `//config:gen-config.yaml`


cmd/merge/main.go
```diff
@@ -59,17 +61,29 @@ func (fm flagMap) Set(s string) error {
...
+	ignoreTeams bool
+}
+
 func main() {
-	orgs := flagMap{}
-	flag.Var(orgs, "org-part", "Each instance adds an org-name=org.yaml part")
+	o := options{orgs: flagMap{}}
+	flag.Var(o.orgs, "org-part", "Each instance adds an org-name=org.yaml part")
+	flag.BoolVar(&o.mergeTeams, "merge-teams", false, "Merge team-name/team.yaml files in each org.yaml dir")
+	flag.BoolVar(&o.ignoreTeams, "ignore-teams", false, "Never configure teams")
```
cblecker: Is this just a safeguard? Looks like this would be the default behaviour.

fejta: Currently if your `org.yaml` defines a team merge will disconfigure it by setting `.Teams = nil`. 

With this change there are three options:
1) Configure whatever is defined in `./org.yaml` (new default behavior)
2) Merge any teams defined by `./foo/teams.yaml` into the `./org.yaml` def (`--merge-teams`)
3) Force `.Teams = nil` (--ignore-teams, current behavior)

None of the `org.yaml` entries set `teams:` so this is a noop, but this prevents someone from deleting all the team members by merging a PR that sets this field.

cblecker: Ahh, got it. Makes perfect sense 👍

PR #279, opened 2018-11-29
Bump dependencies

cblecker:
Brings in peribolos fix, as well as updated dependencies from k8s.io/test-infra


fejta:
+90k -10k... barf

/lgtm

cblecker:
`k8s.io/test-infra/prow/config` started importing `github.com/knative/build` stuff, which then imported client-go and a bunch of other stuff :(

Issue #285, opened 2018-12-04
New repository for tracing mutating admission controller


dashpole:
### New Repo, Staging Repo, or migrate existing
New repository

### Requested name for new repository
mutating-trace-admission-controller

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
piosz, brancz, dashpole

### If not a staging repo, who should have write access
Monkeyanator

### If a new repo, who should be listed as approvers in OWNERS
piosz, brancz, dashpole

### If a new repo, who should be listed in SECURITY_CONTACTS
piosz, brancz, dashpole

### What should the repo description be
a mutating admission controller which enables experimental tracing of kubernetes object lifecycle

### What SIG and subproject does this fall under in sigs.yaml
This is a new subproject for sig-instrumentation called mutating-trace-admission-controller

### Approvals
[Meeting notes from 2018-11-29](https://docs.google.com/document/d/17emKiwJeqfrCsv0NZ2FtyDbenXGtTNCsDEiLbPa7x7Y) say "house mutating webhook for adding trace to an object in kubernetes-sigs"

[Sig-instrumentation charter](https://github.com/kubernetes/community/blob/master/sig-instrumentation/charter.md#subproject-creation) specifies that subproject creation is "By SIG Technical Leads"

### Additional context for request
See https://github.com/Monkeyanator/mutating-trace-admission-controller for @Monkeyanator's current work on the mutating admission controller for tracing.

/assign @brancz @piosz 
for explicit approval

/area github-repo

brancz:
As already discussed in the meeting, this has my approval. This would be our first `kubernetes-sigs/` repo, it looks like we will also need to add members owning the project + sig leads to the kubernetes-sigs org in addition to the repo creation (which doesn't seem to be automated in this repo as opposed to org membership).

dashpole:
@brancz it looks like we need to each open https://github.com/kubernetes/org/issues/new?template=membership.md for that.  @Monkeyenator and I have already done so.

dashpole:
friendly ping to @kubernetes/owners, as this is out of [SLO](https://github.com/kubernetes/community/blob/master/github-management/org-owners-guide.md#slos)

cblecker:
@dashpole Sorry, things have been extra busy around KubeCon. Will review this as soon as I have time.

cblecker:
Thanks @calebamiles!

calebamiles:
[created][]. Please update `sigs.yaml`.

[...]

PR #305, opened 2018-12-13
Add piosz, brancz, dashpole to kubernetes-sigs

calebamiles:
as required to fully implement https://github.com/kubernetes/org/issues/285


cblecker:
@calebamiles per our standards (https://github.com/kubernetes/community/blob/master/github-management/new-membership-procedure.md#processing-the-request) you could rebase your branch, splitting up your commits into one commit per user? They can all be in the same PR, but just one per commit.

PR #335, opened 2019-01-07
Bump org dependencies

cblecker:
Update WORKSPACE and dep dependencies for k8s.io/org. This will bring in some peribolos fixes.


fejta:
How would you feel about my getting this to work with go modules instead (and possibly deleting the vendor dir in the process)?

cblecker:
@fejta This is passing tests now. I'd be totally down with moving to go modules, but this may be complicated with the transitive dependencies on client-go.

I'd still like to get this bump into get the peribolos bug fixes in, and not block my work on peribolos for teams :)

PR #337, opened 2019-01-08
Enable team management for peribolos

### LINKED ISSUE #336: Automate the management of github team membership
/kind design
/sig contributor-experience
/area github

I was unable to find an issue to point to for "we intend to one day automate team membership via peribolos" so now I've gone and created one.

I have used peribolos in the past to mass-create some teams for kubernetes-csi repos, but it wasn't a very friendly experience.  I used a file to generate the teams and populate members/maintainers, but I still had to go through and manually assign ACLs to repos.  

Ultimately I think we want to make this a self-service experience akin to test-infra's config/jobs directory, where files are sharded and OWNERS files allow sigs to control their thing.  But I hazily recall we have discussed in the past the idea of iterating with one massive file per org to start with.

There's also the question of whether we want to disallow team maintainers as a role, and instead just use members + OWNERS files here.  Today I can find no documented policy on who is supposed to be a Team Maintainer and what their responsibilities are.

/assign @cblecker @fejta 
as the two who have the most hands on experience and best chance of enumerating what remains to be done here

cblecker:
🙊!

This might actually work.
fixes #336

/assign @fejta @spiffxp 
/hold

### DIFF
--- a/config/kubernetes-client/org.yaml
+++ b/config/kubernetes-client/org.yaml
@@ -44,3 +44,68 @@ members:
 - webwurst
 - yliaog
 - yue9944882
+teams:
+  Reviewers:
+    description: ""
+    members:
+    - arschles
+    - mbohlool
+    - krabhishek8260
+    - lwander
+    privacy: secret
+  csharp-admins:
+    description: admin access to csharp
+    maintainers:
+    - krabhishek8260
+    members:
+    - brendandburns
+    privacy: closed
+  gen-admins:
+    description: admin access to gen
+    maintainers:
+    - mbohlool
+    members:
+    - brendandburns
+    privacy: closed
+  haskell-admins:
+    description: admin access to haskell
+    maintainers:
+    - mbohlool
+    members:
+    - brendandburns
+    privacy: closed
+  java-admins:
+    description: admin access to java
+    maintainers:
+    - mbohlool
+    members:
+    - brendandburns
+    privacy: closed
+  javascript-admins:
+    description: admin access to javascript
+    maintainers:
+    - mbohlool
+    members:
+    - brendandburns
+    privacy: closed
+  python-admins:
+    description: admin access to python
+    maintainers:
+    - mbohlool
+    members:
+    - yliaog
+    privacy: closed
+  python-base-admins:
+    description: ""
+    maintainers:
+    - mbohlool
+    members:
+    - yliaog
+    privacy: closed
+  ruby-admins:
+    description: admin access to ruby
+    maintainers:
+    - mbohlool
+    members:
+    - brendandburns
+    privacy: closed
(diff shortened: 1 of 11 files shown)


nikhita:
@cblecker :tada: :balloon: 

Two questions:

- Can we move the teams to a `teams.yaml` and merge the configs to avoid bloating `org.yaml`, [like we do for contribex](https://github.com/kubernetes/org/blob/master/config/kubernetes/sig-contributor-experience/teams.yaml)? Or do we intend to move to a `teams.yaml` for each SIG later?

- Out of curiosity, how does the `--dump` and `--fix-teams` come to know that sig-contribex teams should be dumped in its `teams.yaml` file and not append those in `org.yaml`? :thinking:

cblecker:
@nikhita Thanks for feedback!
> Can we move the teams to a teams.yaml and merge the configs to avoid bloating org.yaml, like we do for contribex?

Yes, we can.. it just requires manually trimming out the groups into a new folder, like contribex and apimachinery. I haven't done that yet, because it's a manual process, and I'm currently trying to keep the dump sync'd with what's in GitHub.

> Out of curiosity, how do --dump, --fix-teams and --fix-team-members come to know that sig-contribex teams should be dumped in its teams.yaml file and not append those in org.yaml?

Manually :) I dumped out the list of teams, and then extracted out contribex and apimachinery as examples. I want to get peribolos as the source of truth first, and then we can do it for all the sigs to delegate approve rights.

fejta:
> Manually :) I dumped out the list of teams, and then extracted out contribex and apimachinery as examples. 

However we organize the files https://github.com/kubernetes/org/tree/master/cmd/merge will then stitch them back together into one config.

nikhita:
Manual sounds like a lot of work though :stuck_out_tongue: ....if teams have consistent naming, maybe we could automate it? But I think we should discuss this later on an issue somewhere and not on this PR. Thanks for getting this ready, again! :100: :shipit:

config/kubernetes-client/org.yaml
```diff
@@ -44,3 +44,74 @@ members:
 - webwurst
 - yliaog
 - yue9944882
+teams:
+  Bots:
+    description: Hmm, not working like I hoped...
+    maintainers:
+    - fejta
+    - k8s-ci-robot
+    privacy: secret
```
fejta: LOL

config/kubernetes-client/org.yaml
```diff
@@ -44,3 +44,74 @@ members:
 - webwurst
 - yliaog
 - yue9944882
+teams:
+  Bots:
+    description: Hmm, not working like I hoped...
+    maintainers:
+    - fejta
```
fejta: why am I a bot?! :grin:

cblecker:
@fejta The reason I'm just turning it all on now is we would need a last "sync" from GitHub before we turn it on anyways. I figure I'll do that Friday morning right before merge.

fejta:
/lgtm
/hold
exciting! I guess Friday afternoon is the natural time to make this switch?

cblecker:
/hold cancel
Let's do this! 🚀

PR #346, opened 2019-01-12
Add SIG Release teams

justaugustus:
Fixes: kubernetes/sig-release#241

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

/sig release

PR #350, opened 2019-01-12
Add a test to check if team members are org members

nikhita:
This would be useful to fail early if someone who is added as a team maintainer or member is not an org member. See https://github.com/kubernetes/org/pull/346#pullrequestreview-191923294 for an example.

This PR also validates:

- a user is not mentioned as both an org admin and org member
- a user is not mentioned as both a team maintainer and member
- no duplicates exist in the list of team maintainers and members

/assign @cblecker @fejta

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/config/org"
+	"k8s.io/test-infra/prow/github"
 
 	"github.com/ghodss/yaml"
 )
@@ -63,10 +64,11 @@ func loadOrg(dir string) (*org.Config, error) {
 	return &cfg, nil
 }
 
-func testDuplicates(list []string) error {
+func testDuplicates(list sets.String) error {
 	found := sets.String{}
 	dups := sets.String{}
-	for _, i := range list {
+	all := list.List()
+	for _, i := range all {
 		if found.Has(i) {
 			dups.Insert(i)
 		}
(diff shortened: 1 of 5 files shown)


config/kubernetes-sigs/org.yaml
```diff
@@ -24,7 +24,7 @@ members:
 - animeshsingh
 - ant31
 - apelisse
-- atoms
+- Atoms
```
nikhita:
```
--- FAIL: TestAllOrgs (0.03s)
    --- FAIL: TestAllOrgs/kubernetes-sigs (0.00s)
        config_test.go:169: The following members of team kubespray-maintainers are not org members: Atoms
FAIL
```

config/config_test.go
```diff
@@ -87,6 +87,28 @@ func isSorted(list []string) bool {
 	return sort.StringsAreSorted(items)
 }
 
+// testTeamMembers checks if team maintainers and members are org members.
+// TODO: also ensure that the list is sorted.
+func testTeamMembers(teams map[string]org.Team, orgMembers sets.String) []error {
```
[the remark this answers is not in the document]
nikhita: Done. Also added a check for duplicate names in the lists.

config/kubernetes-sigs/org.yaml
nikhita: Can we export this as well?

nikhita:
> I wonder if this config validation stuff should actually be functions in the prow/config/ package that we export, so that they're reusable.

+1. Can we also make sure that we are sorting before dumping all names? We would be sorted (sorry couldn't resist) once we sort all names and add a test for new ones in the CI, but it might prove useful to others who will use peribolos.

cblecker:
/lgtm
/approve

Thanks so much @nikhita! 🎈

PR #353, opened 2019-01-14
Add danielqsj as a member to sig instrumentation teams

danielqsj:
I'd like to help/participate in reviews and discussions related to sig instrumentation.

/cc @brancz @piosz @fabxc

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -2762,6 +2762,7 @@ teams:
     - zouyee
     - loburm
     - fgrzadkowski
+    - danielqsj
     privacy: closed
   sig-instrumentation-bugs:
     description: ""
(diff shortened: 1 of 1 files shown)


brancz:
I couldn't find an any documentation on a process for this (as in if this is all we need to do), but we're happy to have you!

/lgtm

brancz:
Maybe @spiffxp can tell us :slightly_smiling_face: .

nikhita:
> I couldn't find an any documentation on a process for this (as in if this is all we need to do)

We still need to document process guidelines for this (https://github.com/kubernetes/org/issues/336#issuecomment-453706817), but this should be good enough for now. :)

brancz:
Thanks for clarifying @nikhita! :slightly_smiling_face: Yes sounds like this is all that is needed.

PR #359, opened 2019-01-15
Use go modules instead of dep for dependency management

fejta:
* Use go 1.11
* Update skylib, rules_go, gazelle repos to match test-infra
* Create `//:go` target, as well as `//:update-minor` and `//:update-patch` targets for updating deps.
* Get `kazel` and `peribolos` from `@io_kubernetes_build` and `@io_k8s_test_infra` repos
* Run `go mod init` to create `k8s.io/org` module
* Make hack/update-deps.sh use `go mod` instead of `dep`
  - Runs `go mod tidy && go mod vendor` to manage vendor dir
  - Runs `//:gazelle -- update-repos --from-file=go.mod` to manage `go_repositories()`

/assign @cblecker


cblecker:
I like this. Will review. 🏆 
/hold

techtonik:
Is there a short tutorial to catch up on dependency management in Go and difference between native and `dep` and `go modules`?

fejta:
See https://github.com/golang/go/wiki/Modules

ibrasho:
Can't we remove the vendor/ directory from the repository now?

fejta:
We could delete vendor, although I'd prefer to make that change later. If we delete vendor then that will require bazel to fetch repos when it builds, when runs the risk of being unable to build if a) the repo isn't already cached and b) isn't available for some reason.


Christoph, do we want to go down the route of deleting vendor? Right now bazel will still use bazel to build.

ibrasho:
We can run (athens)[https://github.com/gomods/athens] and configure builders to use if that's an issue.

cblecker:
@fejta So we would still have the bazel cache, but would have to pull otherwise? I'm still trying to tease apart the interaction between bazel and the go modules

fejta:
Right now:
* `go mod tidy` causes golang to update `go.mod` and `go.sum` to the right values, removing any unnecessary imports
* `go mod vendor` causes golang to create the vendor dir
* `bazel run //:gazelle` currently runs in vendor mode, which causes it to add a `//vendor/foo/bar` dep for every `import "foo/bar"`.
* `bazel run //:gazelle -- update-repos --from-file=go.mod` causes it to add `go_repository("foo/bar")` for all the vendored dirs.

We could switch gazelle to run in external mode, which will cause it to add a `@foo/bar` dep for every `import  "foo/bar"`, which will cause it to use the `go_repository()` (aka download and cache the repo) instead of the vendored files

However this has the downside listed above (repo might be inaccessible and uncached)

fejta:
I feel like the safer path is to transition first to having `go mod` manage vendor, and consider how/whether we want to eliminate vendor at a later point

cblecker:
I'd be okay with that if go mods weren't pulling in all these non-go files.. protobufs, shell scripts, testdata, etc.. ugh.

fejta:
Meaning what?
* You'd rather delete vendor and jump immediately to using a cache?
* Stick with dep for now
* Just deal with it for now and wait for `go mod tidy && go mod vendor` to improve in the future?

cblecker:
Would it be awful to do something like this in the update-deps file?
```
find vendor -type f -not -iname "*.go" -not -iname "*.s" -not -iname "LICENSE*" -not -iname "AUTHORS*" -not -iname "CONTRIBUTORS*" -not -iname "LICENSE*" -exec rm {} \;
```
Like.. after the `go mod vendor`, but before the update bazel step?

fejta:
Hey if it passes unit tests sgtm. Let me try it out

fejta:
> Would it be awful to do something like this in the update-deps file?
Done

hack/update-deps.sh
```diff
@@ -29,20 +29,22 @@ set -o xtrace
...
-rm -rf vendor/github.com/jteeuwen/go-bindata/testdata
-rm -rf vendor/github.com/docker/docker/contrib
+prune-vendor() {
+  find vendor -type f \
+    -not -iname "*.go" \
+    -not -iname "*.s" \
+    -not -iname "LICENSE*" \
+    -not -iname "AUTHORS*" \
+    -not -iname "CONTRIBUTORS*" \
+    -not -iname "LICENSE*" \
```
cblecker: my bad. I duplicated this one by accident.

fejta: ... or maybe this was a test that I failed horribly? :smiling_imp: 

Also keeping .proto, .h, .c files

fejta:
Woo! Tests are a passin'

cblecker:
/lgtm
/hold

I'm good with this, but I'd like to wait until after our team sync has run a few times before merging.. just so we can separate out which change broke something if something breaks.

fejta:
Recommend unholding this so we can see it run with the higher token limit  https://github.com/kubernetes/test-infra/pull/10803

fejta:
/hold cancel
Got the green light from Sir Christoph

PR #376, opened 2019-01-17
config/kubernetes: Add mxinden to sig-instrumentation

mxinden:
Following https://github.com/kubernetes/org/pull/353 as an example, I would like to help in sig-instrumentation. Hence being added here would be great.

Let me know if this is the wrong way to approach this.

//CC @brancz


brancz:
/lgtm

Happy to finally have you officially be part of sig-instrumentation! :slightly_smiling_face:

PR #387, opened 2019-01-19
Add 1.14 release team members to sig-release github teams

spiffxp:
Add 1.14 release team members to the kubernetes/milestone-maintainers team
 - I expect that at a minimum everyone in the 1.14 release team roles should have /milestone privileges.
 - Role shadows I am willing to let roles decide for themselves.

Add 1.14 release team lead + shadows to the kubernetes/kubernetes-release-managers team:
- solely for "just in case" purposes, and you can totally convince me we should be revoked down the line, just following the description of the team
- provide coverage in case branch manager doesn't have availability

Also, having just gone through the exercise of doing a commit per user, I did not enjoy it. I know that's the process we use for adjusting org membership, but it doesn't feel like it scales very well. But I did it this time just to try it out

### DIFF
--- a/config/kubernetes/sig-release/teams.yaml
+++ b/config/kubernetes/sig-release/teams.yaml
@@ -87,6 +87,7 @@ teams:
     - calebamiles
     - justaugustus
     - tpepper # 1.13 PRM
+    - spiffxp # 1.14 RT Lead
     members:
     - aleksandra-malinowska # 1.13 PRM
     - feiskyer # 1.12 PRM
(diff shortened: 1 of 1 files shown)


spiffxp:
(I also sorta wish the comments followed the keys of kubernetes/community/sigs.yaml and the directories in kubernetes/sig-release/release-team... all-lowercase-with-dashes, but that can be a separate PR if I'm still itchy over it later)

justaugustus:
Yeah, the user per commit thing is not great.
(cc: @cblecker)

/lgtm
/approve

PR #405, opened 2019-01-23
Add dims to k8s.io/utils

dims:
Change-Id: Ib1c453e2889fa9afe16f0afdb16d81be343922b8

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1915,6 +1915,7 @@ teams:
     description: ""
     maintainers:
     - apelisse
+    - dims
     members:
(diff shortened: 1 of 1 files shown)


config/kubernetes/org.yaml
```diff
@@ -1915,29 +1915,35 @@ teams:
...
     description: ""
     maintainers:
     - apelisse
+    - dims
     members:
     - mengqiy
     - monopole
     - pwittrock
     - thockin
+    - dims
```
nikhita: If you add yourself as a maintainer, you don't need to add as a member (peribolos will throw an error)

PR #406, opened 2019-01-23
Add bash-firefighters team

cblecker:
:bash_fire:

@kubernetes/sig-testing anyone else want in on this?
cc: @fejta @BenTheElder @cjwagner @ixdy @stevekuznetsov

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -811,6 +811,19 @@ teams:
     - MaciekPytel
     - piosz
     privacy: closed
+  bash-firefighters:
+    description: Folks with expertise in bash reviews
+    maintainers:
+    - cblecker
+    - fejta
+    - spiffxp
+    members:
+    - BenTheElder
+    - cjwagner
+    - ixdy
+    - stevekuznetsov
+    - sttts
+    privacy: closed
   bots:
     description: Bot Service Accounts in the Kubernetes org
     maintainers:


cblecker:
by popular demand https://github.com/kubernetes/repo-infra/issues/97#issuecomment-456841978

ixdy:
can't tell if serious

cblecker:
This is a legitimate PR that will create a legitimate team haha

BenTheElder:
Looking forward to the bash BOF at the next KubeCon 😉

stevekuznetsov:
:bash_fire"

nikhita:
I think @sttts should be on the team as well (sorry Stefan! :laughing: )

spiffxp:
Does it need to be sig-testing? I feel like there are many dirs that have bash that are not just under sig-testing's provenance

spiffxp:
Common dirs that need this expertise:
- kubernetetes/repo-infra - a contribex subproject
- kubernetes/kubernetes/hack - doesn't actually belong to any one sig
- kubernetes/kubernetes/build - same (but probably should?)

config/kubernetes/sig-testing/teams.yaml
```diff
@@ -37,3 +37,14 @@ teams:
     - cjwagner
     - mkumatag
     privacy: closed
+  sig-testing-bash-reviews:
```
spiffxp: If you drop sig-testing from the name, then I'm fine with it, though I still think this is a volunteer effort at best and may not improve the situation.  All of these folks are pretty oversubscribed as it is

config/kubernetes/org.yaml
```diff
@@ -811,6 +811,18 @@ teams:
...
+  bash-firefighters:
+    description: "Folks with expertise in bash reviews"
+    maintainers:
+    - cblecker
+    - fejta
+    members:
+    - BenTheElder
+    - cjwagner
+    - ixdy
+    - stevekuznetsov
```
spiffxp: _...sighs_.... and my axe

put me in coach

PR #429, opened 2019-01-29
Move test-infra-* GitHub teams to sig-testing

spiffxp:


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -37,3 +37,61 @@ teams:
     - cjwagner
     - mkumatag
     privacy: closed
+  test-infra-admins:
+    description: Administrators to test-infra repo
```
stevekuznetsov: Should I be in this one?

config/kubernetes/sig-testing/teams.yaml
```diff
@@ -37,3 +37,61 @@ teams:
...
+    - Random-Liu
+    - rmmh
+    - shyamjvs
+    - stevekuznetsov
+    - sttts
+    - wojtek-t
+    - wonderfly
+    - zmerlynn
+    privacy: closed
+  test-infra-reviewers:
```
stevekuznetsov: What's the use of this?

config/kubernetes/sig-testing/teams.yaml
BenTheElder: this was just a move of the existing lists above to scope them into the SIG. there should be a follow up to clean up the lists.

BenTheElder: all of these teams are copied from what is in github, and are therefore almost undoubtedly out of date.

stevekuznetsov: SGTM

PR #435, opened 2019-01-30
Updated OWNERS files to include link to docs

rlenferink:
As per PR #431 all the other OWNER files are updated to include the link to the documentation as well now.

/assign @cblecker


rlenferink:
Well, it seems that our bot is doing its job with adding labels 😄

PR #538, opened 2019-02-27
Rename PST team to PSC

nikhita:
/assign @cblecker @fejta 

Ref: https://github.com/kubernetes/steering/issues/89

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1863,7 +1863,7 @@ teams:
     - gmarek
     - wojtek-t
     privacy: closed
-  product-security-team:
+  product-security-committee:
     description: Please report security issues to https://kubernetes.io/security
     members:
     - cjcullen
@@ -1872,6 +1872,8 @@ teams:
     - philips
     - tallclair
     privacy: closed
+    previously:
+    - product-security-team
   repo-infra-maintainers:
     description: ""
     members:


config/kubernetes/org.yaml
```diff
@@ -1863,7 +1863,7 @@ teams:
     - gmarek
     - wojtek-t
     privacy: closed
-  product-security-team:
+  product-security-committee:
```
nikhita: @cblecker @fejta peribolos should delete the old team and create a new team for PSC, right? Can you sanity check this once?

nikhita:
fyi @kubernetes/product-security-team

config/kubernetes/org.yaml
nikhita: Explicitly holding for this comment.

/hold

cblecker: If you do something like this with the previous name, it will rename the team, instead of delete/recreate:
https://github.com/kubernetes/org/blob/9794f30916f026dc3b03bb7af78f7ec2bea90bb5/config/kubernetes/sig-pm/teams.yaml#L10-L11

nikhita: Ah right, I was pretty sure I was forgetting something. :)

Done. :+1:

nikhita:
Cancelling the hold since https://github.com/kubernetes/community/pull/3311 has merged.
/hold cancel

PR #534, opened 2019-02-27
Add teams for French localization

nikhita:
Fixes https://github.com/kubernetes/org/issues/532

/assign @zacharysarah

Issue #532, opened 2019-02-27
Request: new org teams


zacharysarah:
### Organization or repository

Kubernetes org

### Users affected

Members of the `sig-docs-fr-*` aliases in [k/website/OWNERS_ALIASES](https://github.com/kubernetes/website/blob/master/OWNERS_ALIASES#L140-L163)

### Describe the issue

Heya! 👋 Now that docs have an active French localization, we need to add two org teams for splat reviews and mentions:
- [ ] `sig-docs-fr-owners`
- [ ] `sig-docs-fr-reviews`

Team members should match their [respective aliases](https://github.com/kubernetes/website/blob/master/OWNERS_ALIASES#L140-L163) in k/website.

/area github-membership
/priority important-soon
/sig docs

PR #546, opened 2019-02-28
Add sig docs it teams and update fr teams

rlenferink:
PR is adding the Italian localization teams. The members are already in the appropriate OWNERS files => [kubernetes/website/OWNERS_ALIASES](https://github.com/kubernetes/website/blob/12036aa614ff0e4ac5bdfce102642b7e5e4cb2b5/OWNERS_ALIASES#L166)

Next to that some small improvements
- Ignoring `.idea` files when opening the project with a Jetbrains editor
- Swapped `sig-docs-fr-reviews` and `sig-docs-fr-owners` to follow the rest of the file: first owners, then review teams
- Fixed indentation for `sig-docs-ko-reviews`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *~
 /bazel-*
 .DS_Store
+.idea/
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -102,7 +102,6 @@ teams:
     - ClaudiaJKang
     - ianychoi
     privacy: closed
-    teams:
   sig-docs-ko-reviews:
     description: Reviews for Korean docs PRs
     maintainers:
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -36,6 +36,7 @@ teams:
     - awkif
     - jygastaud
     - lledru
+    - oussemos
     - perriea
     - rbenzair
     - rekcah78
@@ -50,6 +51,7 @@ teams:
     - awkif
     - jygastaud
     - lledru
+    - oussemos
     - perriea
     - rbenzair
     - rekcah78
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -55,6 +55,22 @@ teams:
     - smana
     - yastij
     privacy: closed
+  sig-docs-it-owners:
+    description: Owners for Italian content
+    maintainers:
+    - rlenferink
+    members:
+    - lledru
+    - micheleberardi
+    privacy: closed
+  sig-docs-it-reviews:
+    description: Review PRs for Italian content
+    maintainers:
+    - rlenferink
+    members:
+    - lledru
+    - micheleberardi
+    privacy: closed
   sig-docs-ja-owners:
     description: sig-docs-ja-owners
     maintainers:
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 *~
 /bazel-*
 .DS_Store
+.idea/
```
nikhita: For the record, I'm okay with this because this is present in k/k's [gitignore](https://github.com/kubernetes/kubernetes/blob/983cf51c53431170e11d1d41a820ee6bd7bd4bcc/.gitignore#L13) and afaik generally needed for folks working with JetBrains IDEs.

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -85,14 +101,14 @@ teams:
...
     teams:
-      sig-docs-ko-reviews:
-        description: Reviews for Korean docs PRs
-        maintainers:
-        - ClaudiaJKang
-        - gochist
-        members:
-        - ianychoi
-        privacy: closed
+  sig-docs-ko-reviews:
```
nikhita: Thanks for fixing this, @rlenferink!

@cblecker @fejta We need `ko-reviewers` to _not_ be nested under `ko-owners` because child teams inherit parents' teams permissions iirc. Can you sanity check that changing the indentation will remove this as a child team and not have any side effects?

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -29,8 +29,8 @@ teams:
     - xiangpengzhao
     - zhangxiaoyu-zidif
     privacy: closed
-  sig-docs-fr-reviews:
-    description: Review PRs for French content
+  sig-docs-fr-owners:
```
nikhita: @rlenferink While we are here, can you add @abuisine to fr-owners and fr-reviewers?

Ref: https://github.com/kubernetes/org/pull/534#issuecomment-468235782

.gitignore
rlenferink: That's why I was adding this (and only this) line 😃

config/kubernetes/sig-docs/teams.yaml
rlenferink: Done: https://github.com/kubernetes/org/pull/546/commits/1878a0e835f8ff696493f1548106c2100126b685

rlenferink:
@nikhita thanks for your feedback!

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -42,9 +43,10 @@ teams:
     - smana
     - yastij
     privacy: closed
-  sig-docs-fr-owners:
-    description: Owners for French content
+  sig-docs-fr-reviews:
```
nikhita: @rlenferink sorry for the extra trouble :see_no_evil: ...but could you also add @oussemos to fr-owners and fr-reviews?

Ref: https://github.com/kubernetes/website/pull/12708 and https://github.com/kubernetes/org/issues/532

rlenferink: Not a problem at all. I’ve updated this PR to include @oussemos as well ;)

config/kubernetes/sig-docs/teams.yaml
fejta: I'm not sure about the no side effects part... Worst case it will delete the team ID and then recreate a new team and add the same people to it.

rlenferink:
Added https://github.com/kubernetes/org/pull/546/commits/4ad75502e9ec5f1bd6a8a921ac431054034b6b2a as suggested by #563

PR #552, opened 2019-03-02
Add rjbez17 to conformance wg members

rjbez17:


spiffxp:
/approve
the list of usernames needs to be alphabetized, I'll /lgtm when that's fixed

spiffxp:
/ok-to-test
/lgtm
(I suspect this will fail until you're added as an org member)

PR #558, opened 2019-03-04
Add publishing-bot-reviewers team

nikhita:
Adds a team so that we can use it to cc folks for reviews/issues related to the publishing-bot.

Ref: https://github.com/kubernetes/publishing-bot/blob/master/OWNERS

/cc @sttts @dims @mfojtik @caesarxuchao

### DIFF
--- a/config/kubernetes/sig-release/teams.yaml
+++ b/config/kubernetes/sig-release/teams.yaml
@@ -179,3 +179,13 @@ teams:
         - justaugustus
         - tpepper
         privacy: closed
+  publishing-bot-reviewers:
+    description: Reviews for publishing-bot
+    maintainers:
+    - nikhita
+    members:
+    - caesarxuchao
+    - dims
+    - mfojtik
+    - sttts
+    privacy: closed


config/kubernetes/sig-release/teams.yaml
```diff
@@ -178,3 +178,13 @@ teams:
         - justaugustus
         - tpepper
         privacy: closed
+  publishing-bot-reviewers:
+    description: Reviews for publishing-bot
+    maintainers:
+    - caesarxuchao
+    - nikhita
```
nikhita: maintainers = approvers
members = reviewers

I've added myself here because the GH API requires that org owners be in the maintainers list. They can't be added as members. :/

nikhita:
/hold

Want to specifically get lgtm from @sttts

config/kubernetes/sig-release/teams.yaml
rlenferink: While we're converting maintainers -> members (https://github.com/kubernetes/org/issues/563), it may be good to add `caesarxuchao` to the members right away instead of the maintainers section 😄 (org owners aside)

nikhita: Thanks @rlenferink. I've updated as per https://github.com/kubernetes/org/issues/563.

Issue #308, opened 2018-12-14
Please create repo kubernetes-sigs/etcdadm


justinsb:
### New Repo, Staging Repo, or migrate existing
New Repo

### Requested name for new repository
etcdadm

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
justinsb, dlipovetsky & sig-cluster-lifecycle chairs (roberthbailey, luxas, timothysc)

### If not a staging repo, who should have write access
justinsb, dlipovetsky

### If a new repo, who should be listed as approvers in OWNERS
justinsb, dlipovetsky

### If a new repo, who should be listed in SECURITY_CONTACTS
justinsb, dlipovetsky

### What should the repo description be
tooling & automation for kubernetes etcd clusters

### What SIG and subproject does this fall under in sigs.yaml
etcdadm is a new subproject of sig-cluster-lifecycle

### Approvals
https://github.com/kubernetes/community/pull/2835

### Additional context for request
n/a

/area github-repo

cblecker:
I can’t tell clearly from the KEP.. are we forking an existing project?

jberkus:
@cblecker yes.  I believe this is the current canonical source: https://github.com/platform9/etcdadm

cblecker:
@justinsb can you confirm? rules for importing/donating code is different than a new repo with newly created code.

dlipovetsky:
@cblecker Platform9 would like to donate the code. We can transfer the repo the kubernetes-sigs org, or duplicate the repo (and outstanding issues).

cblecker:
Can you please review the rules for donated repositories, and speak to each of the requirements?: https://github.com/kubernetes/community/blob/master/github-management/kubernetes-repositories.md#rules-for-new-repositories

also, please add me as an admin to the current repo

dlipovetsky:
@cblecker Absolutely.
- I confirmed that all [contributors](https://github.com/platform9/etcdadm/graphs/contributors) have signed the CNCF CLA.
- Boilerplate text has been [updated](https://github.com/platform9/etcdadm/pull/83) to attribute copyright to "The etcdadm authors"
- Dependencies have MIT, AL2.0, or BSD licenses [0], which I understand are all acceptable.

Although it does not appear in the rules, I realize that--where possible--the use of the same dependencies as k/k is important, too; @justinsb_ was kind enough to submit one patch getting etcdadm closer to this goal. 

[0]
```
$ licenses github.com/platform9/etcdadm
[...]

Issue #563, opened 2019-03-05
Convert GitHub team maintainers to members


spiffxp:
Now that we automate the management of GitHub teams (https://github.com/kubernetes/org/issues/336), it is confusing when people try to use GitHub's UI to manually manage teams, only to have their changes reverted (eg: https://github.com/kubernetes/org/issues/559)

While we wait for docs (https://github.com/kubernetes/community/issues/3102) I believe we should move everyone to being a regular team member, and encourage use of PR's against this repo for GitHub team management

That said, one fun little wrinkle I discovered in https://github.com/kubernetes/org/pull/522 is that GitHub Owners need to remain team maintainers

/kind cleanup
/priority important-soon

spiffxp:
[Notified kubernetes-dev@](https://groups.google.com/forum/#!topic/kubernetes-dev/WtvHtt2lR1w)

nikhita:
I'll update the docs and config.

/assign

rlenferink:
> I'll update the docs and config.
> 
> /assign

While we're at it with PR https://github.com/kubernetes/org/pull/546, should I update the other `sig-docs` teams as well to 'convert' maintainers to members?

nikhita:
> @nikhita While we're at it with PR #546, should I update the other sig-docs teams as well to 'convert' maintainers to members?

Sure, that would be good. :)

nikhita:
Created https://github.com/kubernetes/org/pull/566 for the config part.

I'll add the docs later this week.

nikhita:
Created https://github.com/kubernetes/community/pull/3368 for the docs.

nikhita:
The config part has now merged: https://github.com/kubernetes/org/pull/566. All existing (non-admin) team maintainers have been converted to members and a test has been added to check that non-admins are not listed as maintainers.

Also notified k-dev that this is done: https://groups.google.com/d/msg/kubernetes-dev/WtvHtt2lR1w/TMiaTjESCAAJ

cblecker:
Take two: https://github.com/kubernetes/org/pull/849

justaugustus:
This is now complete, right?

nikhita:
Yes! :tada: 

/close

PR #569, opened 2019-03-06
Add dlipovetsky to etcdadm teams

cblecker:
ref https://github.com/kubernetes/org/issues/308#issuecomment-458360650


config/kubernetes-sigs/org.yaml
```diff
@@ -425,11 +425,13 @@ teams:
   etcdadm-admins:
     description: admin access to etcdadm
     maintainers:
+    - dlipovetsky
```
nikhita: We want to migrate maintainers to members https://github.com/kubernetes/org/issues/563, but let's get this merged and I'll fix it in #566.

PR #591, opened 2019-03-11
Revert migrate maintainers to members for kubernetes-incubator

nikhita:
Peribolos [fails](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-org-peribolos/201) with a 403 when it tries to update memberships for certain teams in kubernetes-incubator. This temporarily unblocks peribolos so other changes can be synced.

```
{"client":"github","component":"peribolos","level":"info","msg":"UpdateTeamMembership(2113629, euank, false)","time":"2019-03-11T03:17:04Z"}
{"component":"peribolos","error":"is the account using at least one of the following oauth scopes?: admin:org, repo","level":"warning","msg":"UpdateTeamMembership(2113629, euank, false) failed","time":"2019-03-11T03:17:07Z"}
{"component":"peribolos","level":"fatal","msg":"Configuration failed: failed to configure kubernetes-incubator teams: failed to update maintainers-rktlet members: 1 errors: [is the account using at least one of the following oauth scopes?: admin:org, repo]","time":"2019-03-11T03:17:07Z"}
```

The relevant diff from https://github.com/kubernetes/org/pull/566: https://github.com/kubernetes/org/pull/566/files#diff-3b95bdd66ee80ec3ebbed25503db3376

Note: I am only reverting the change for k-incubator here because teams in other orgs have been synced (some teams in kuberntes-incubator have also been synced but it should be ok to have that change reverted). I'm also not completely sure that this will fix the error, but it seems the best solution to me right now. I have created https://github.com/kubernetes/test-infra/pull/11720 to help debug better in the future. 

For some more context, the relevant slack discussion: https://kubernetes.slack.com/archives/C1TU9EB9S/p1552055216157000

/hold
Holding for lgtm from @spiffxp or @fejta 

/cc @spiffxp @mrbobbytables @justaugustus @rlenferink @idealhack 
/assign @spiffxp @fejta

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -108,11 +108,6 @@ func testTeamMembers(teams map[string]org.Team, admins sets.String, orgMembers s
 		teamMaintainers = normalize(teamMaintainers)
 		teamMembers = normalize(teamMembers)
 
-		// check for non-admins in maintainers list
-		if nonAdminMaintainers := teamMaintainers.Difference(admins); len(nonAdminMaintainers) > 0 {
-			errs = append(errs, fmt.Errorf("The team %s in org %s has non-admins listed as maintainers; these users should be in the members list instead: %s", teamName, orgName, strings.Join(nonAdminMaintainers.List(), ",")))
-		}
-
 		// check for users in both maintainers and members
 		if both := teamMaintainers.Intersection(teamMembers); len(both) > 0 {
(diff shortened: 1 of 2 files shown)


nikhita:
Also, a quick note that this doesn't occur due to rate limiting problems. We check for the scopes _after_ checking for exceeding the rate limiting or hitting abuse detection.

PR #592, opened 2019-03-11
Revert 'Revert migrate maintainers to members for k-incubator'

nikhita:
Reverts #591 

#591 fixed peribolos and it's now green...but we still need to migrate maintainers to members for k-incubator.

/cc @spiffxp @mrbobbytables @justaugustus @rlenferink @idealhack 
/assign @spiffxp @fejta 
/hold

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -108,6 +108,11 @@ func testTeamMembers(teams map[string]org.Team, admins sets.String, orgMembers s
 		teamMaintainers = normalize(teamMaintainers)
 		teamMembers = normalize(teamMembers)
 
+		// check for non-admins in maintainers list
+		if nonAdminMaintainers := teamMaintainers.Difference(admins); len(nonAdminMaintainers) > 0 {
+			errs = append(errs, fmt.Errorf("The team %s in org %s has non-admins listed as maintainers; these users should be in the members list instead: %s", teamName, orgName, strings.Join(nonAdminMaintainers.List(), ",")))
+		}
+
 		// check for users in both maintainers and members
 		if both := teamMaintainers.Intersection(teamMembers); len(both) > 0 {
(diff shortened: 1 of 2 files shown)


fejta:
If reverting #591 is what fixed peribolos, won't adding it back break peribolos again?

justaugustus:
> If reverting #591 is what fixed peribolos, won't adding it back break peribolos again?

Yeah, I'm also curious as to what's changed.

nikhita:
Yeah, I am not entirely sure _why_ we were hitting a 403 in the first place. Note that we weren't hitting the token limit because this error is displayed _after_ checking if we've exceeded the rate limit.

It also looked like we were getting tripped on the same team each time. I was thinking that we could make changes to all teams, but **except the problematic team** and try syncing the problematic team in another PR. Any other suggestions?

justaugustus:
Ahhh, the offending team is not in this PR? If that's the case, then SGTM!

rlenferink:
Trying to remove the problematic team and see how it behaves may be good to get started with indeed.

@nikhita can you update your PR to exclude changes to the [maintainers-rktlet team](https://github.com/kubernetes/org/pull/592/files#diff-3b95bdd66ee80ec3ebbed25503db3376L586), as that's the team [failing](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-org-peribolos/201)?

nikhita:
(I'm semi-afk today, will do that tomorrow :grimacing: )

rlenferink:
np, take your time 🙂

fejta:
This still alive?

spiffxp:
needs rebase

cblecker:
Closing in favour of https://github.com/kubernetes/org/pull/849

PR #597, opened 2019-03-13
Clean up group membership for roberthbailey

roberthbailey:
This is primarily to reduce email load. 

/assign @spiffxp

### DIFF
--- a/config/kubernetes-incubator/org.yaml
+++ b/config/kubernetes-incubator/org.yaml
@@ -639,7 +639,6 @@ teams:
     - lavalamp
     - philips
     - piosz
-    - roberthbailey
     - sarahnovotny
     - smarterclayton
     - spxtr
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -274,7 +274,6 @@ teams:
     members:
     - kris-nova
     - krousey
-    - roberthbailey
     - timothysc
     privacy: closed
   cluster-api-maintainers:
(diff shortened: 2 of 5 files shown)


OWNERS_ALIASES
```diff
@@ -39,7 +39,6 @@ aliases:
     - hogepodge
     - jagosan
   sig-cluster-lifecycle-leads:
-    - roberthbailey
```
nikhita: The OWNERS_ALIASES file here is supposed to be a mirror/copy of the one in the community repo so I wouldn't remove it right away. :)

https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES#L45

config/kubernetes-incubator/org.yaml
```diff
@@ -639,7 +639,6 @@ teams:
     - lavalamp
     - philips
     - piosz
-    - roberthbailey
```
nikhita: This will remove you as a member for k-incubator. Staying as a member shouldn't affect email notifications so just want to double check this. :)

OWNERS_ALIASES
roberthbailey: Roger. I'll undo this one for now.

config/kubernetes-incubator/org.yaml
roberthbailey: This removes me as a member of the group that owns the incubator org (which seems appropriate). I'm still a member of the org itself (line 137).

nikhita: Ah my bad, the github diff view confused me.

PR #602, opened 2019-03-14
Remove owners team in kubernetes-incubator org

nikhita:
The [GitHub Admin team](https://github.com/kubernetes/community/tree/master/github-management#github-administration-team) is responsible for having admin privileges across all Kubernetes orgs.

/assign @idvoretskyi @spiffxp

### DIFF
--- a/config/kubernetes-incubator/org.yaml
+++ b/config/kubernetes-incubator/org.yaml
@@ -626,24 +626,6 @@ teams:
     - squat
     - thockin
     privacy: closed
-  owners:
-    description: owners of the kubernetes-incubator org
-    maintainers:
-    - calebamiles
-    - spiffxp
-    members:
-    - alex-mohr
-    - bgrant0607
-    - ConnorDoyle
-    - eparis
-    - lavalamp
-    - philips
-    - piosz
-    - sarahnovotny
-    - smarterclayton
-    - spxtr
-    - thockin
-    privacy: closed
   reviewers-cri-containerd:
     description: reviewers of cri-containerd repo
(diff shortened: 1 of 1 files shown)


nikhita:
No, I just noticed this because of https://github.com/kubernetes/org/pull/597#discussion_r265344567.

We have a kubernetes/owners@ team today. So having another  `owners` team in kubernetes-incubator is confusing...especially since they don't really have admin rights on the org anyway. This just looks like cruft from the early days of the org.

rlenferink:
Isn't this used for members of `kubernetes-incubator` who are not a member of the `kubernetes` org and therefore cannot ping `kubernetes/owners`?

I agree though that it may be better to have this removed, as it easily leads to inconsistency of team members.

/lgtm
/hold
Holding till we know @idvoretskyi his `lgtm` as well ;)

spiffxp:
I would either remove this, or reconcile with kubernetes/owners (which might then mean you should ensure there's a duplicate owners team in all sigs, and then that's a lot of copy pasta, so maybe we'd want to lint that or generate it and.... so, neither path is perfect)

Personally I haven't found the team used much, and we repeatedly tell folks github notifcations are not an effective mechanism for getting a response.  So I'd opt for deleting it.

spiffxp:
/lgtm
(since this deletes and that's what I'm opting for)

nikhita:
If we come across a scenario where we find having an `owners` team is helpful in other orgs, let's revisit this. :)

Ihor, can you cancel the hold if this lgty?

PR #624, opened 2019-03-18
Add community-milestone-maintainers team

nikhita:
For https://github.com/kubernetes/community/issues/3443

This team will have the power to use the `/milestone` command in the community repo. Membership in this team requires an entry in an OWNERS file (reviewer or approver) for a contribex subproject in the community repo.

Once this merges and the team is created, I'll add the team here: https://github.com/kubernetes/test-infra/blob/b9778912cf2e162c9f2b005cdff2a48be7f4a593/prow/plugins.yaml#L214.

/hold
/assign @parispittman

### DIFF
--- a/config/kubernetes/sig-contributor-experience/teams.yaml
+++ b/config/kubernetes/sig-contributor-experience/teams.yaml
@@ -26,6 +26,30 @@ teams:
     - Phillels
     - thockin
     privacy: closed
+  community-milestone-maintainers:
+    description: Contributors who can use `/milestone` in the community repo. Defined by an entry in an OWNERS file for a contribex subproject.
+    maintainers:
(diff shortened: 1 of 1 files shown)


spiffxp:
/lgtm
FWIW on the test-infra changes, the repo key does not need to be `'quoted'`, we only needed that for the default `''` empty string.

AFAIK that plugin doesn't support hierarchical/overlapping/multiple teams (PR's welcome etc), so your intended change will result in `kuberentes-milestone-maintainers` losing `/milestone` privileges on `kubernetes/community`.  I'm in favor of this, just being clear on the consequences.

parispittman:
/lgtm
/approve

thank you! 🚀

PR #628, opened 2019-03-19
Add milestone team, clean up teams in general

zacharysarah:
For https://github.com/kubernetes/test-infra/issues/11687#issuecomment-474388190.

This PR specifies who can use `/milestone` in the website repo. 

Membership in this team requires an entry in a `sig-docs-**-owners` alias for a localization project in k/website. (See https://github.com/kubernetes/website/pull/13287) Note that `website-milestone-maintainers` imports membership by reference to language owner groups.

This PR also updates sig-docs team membership based on `k/website/OWNERS_ALIASES`.

### DIFF
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -1,26 +1,31 @@
 teams:
-  sig-docs-en-owners:
-    description: ""
+  sig-docs-en-owners: 
+    description: English content admins 
     maintainers:
-    - Bradamant3
-    - chenopis
+    - bradamant3
+    - jaredbhatti
     - zacharysarah
     members:
     - bradtopol
+    - chenopis
     - cody-clark
-    - jaredbhatti
+    - jimangel
     - kbarnard10
     - mistyhacks
+    - rajakavitha1
     - ryanmcginnis
     - steveperry-53
     - stewart-yu
     - tengqm
     - tfogo
+    - zhangxiaoyu-zidif
     - zparnold
     privacy: closed
   sig-docs-en-reviews:
-    description: Review PRs for English content
+    description: PR reviews for English content
     maintainers:
+    - Bradamant3
+    - jaredbhatti
     - zacharysarah
     members:
     - jimangel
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-docs/teams.yaml
```diff
@@ -156,25 +175,36 @@ teams:
...
     - idealhack
     - pigletfly
-    - tengqm
-    - xiangpengzhao
-    - zhangxiaoyu-zidif
     privacy: closed
+  website-milestone-maintainers:
+    description: Contributors who can use `/milestone` in the website repo. Defined in https://github.com/kubernetes/org/blob/master/config/kubernetes/sig-docs/teams.yaml.
+    members:
+    - sig-docs-en-owners
```
nikhita: Unfortunately, teams can't be nested like this :grimacing: 

You'll either need to add this as a "parent team" to the `sig-docs-*` teams like https://github.com/kubernetes/org/blob/master/config/kubernetes/sig-cloud-provider/teams.yaml#L18 or expand all teams and list out members individually.

While the latter seems cumbersome, I'd mostly suggest that because GitHub says that (ref https://help.github.com/en/articles/about-teams#nested-teams):

> Child teams inherit the parent's access permissions, simplifying permissions management for large groups. Members of child teams also receive notifications when the parent team is @mentioned, simplifying communication with multiple groups of people.

I'm not sure if this is a cause for concern here but it _may_ lead to unexpected things with respect to permissions in the future.

zacharysarah: Understood! Listing folks out will be cumbersome for obvious reasons, but better safe than sorry.

zacharysarah:
@nikhita 👋 This PR is green now and ready to merge. 💚

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -156,25 +175,69 @@ teams:
...
+    - tengqm
+    - xiangpengzhao
+    - zhangxiaoyu-zidif
     members:
     - chenrui333
     - idealhack
     - pigletfly
+    privacy: closed
+  website-milestone-maintainers:
+    description: Contributors who can use `/milestone` in the website repo. Defined in https://github.com/kubernetes/org/blob/master/config/kubernetes/sig-docs/teams.yaml.
```
nikhita: Since the link points to this file, maybe we drop it? :)

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -1,151 +1,170 @@
 teams:
-  sig-docs-en-owners:
-    description: ""
+  sig-docs-en-owners: 
+    description: English content admins 
+    maintainers:
+    - bradamant3
```
nikhita: All users need to defined as `members` instead of `maintainers`. Ref: #563 

I have a PR to add a test for it https://github.com/kubernetes/org/pull/592, so this should be caught earlier in the future. :grimacing:

nikhita:
@zacharysarah added a couple of comments, but this looks good to go otherwise!

nikhita:
Oh and one last thing...could you also sqaush the commits? :)

zacharysarah:
@nikhita Two of the commits in this PR are merge commits and won't rebase properly for squashing. I'm going to close this PR and start fresh with a new one.

Issue #630, opened 2019-03-20
New repository: addon-operators


dholbach:
### New Repo, Staging Repo, or migrate existing
new repository

### Requested name for new repository
addon-operators

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
@justinsb 

### If not a staging repo, who should have write access
@justinsb 

### If a new repo, who should be listed as approvers in OWNERS
@justinsb 

### If a new repo, who should be listed in SECURITY_CONTACTS
@justinsb 

### What should the repo description be
The cluster-addons sub-project was set up in [the 2019-03-12 meeting of SIG Cluster Lifecycle](https://docs.google.com/document/d/1Gmc7LyCIL_148a9Tft7pdhdee0NBHdOfHS1SAF0duI4/edit#heading=h.xhid2p5wz968)

* Slack: https://kubernetes.slack.com/messages/CH21PG8SW
* KEP: https://github.com/kubernetes/enhancements/pull/746

### What SIG and subproject does this fall under in sigs.yaml
This is a new subproject for sig-cluster-lifecycle called cluster-addons

### Approvals
 - Meeting minutes 2019-03-12: https://docs.google.com/document/d/1Gmc7LyCIL_148a9Tft7pdhdee0NBHdOfHS1SAF0duI4/edit#heading=h.xhid2p5wz968
 - Decision on name: https://kubernetes.slack.com/archives/C13J86Z63/p1552667179275500
 - Decision on repo name: https://kubernetes.slack.com/archives/C13J86Z63/p1552667403278000

/cc @luxas @timothysc

luxas:
:+1:

nikhita:
Created https://github.com/kubernetes-sigs/addon-operators :tada: 

Once the postsubmit runs for https://github.com/kubernetes/org/pull/631, the teams will be created on GitHub. I will then assign the teams appropriate access to the repo. :)

@dholbach could you create a PR to add this new subproject to sigs.yaml in the community repo?

Once the teams are assigned access + the above PR gets merged, we can close this issue.

nikhita:
Oh also...what should be the one-line description for the repo?

nikhita:
> I will then assign the teams appropriate access to the repo. 

This is done.

dholbach:
> Oh also...what should be the one-line description for the repo?

"Addon operators for Kubernetes clusters"?

Not very witty, but let's add it for now and move on from there.

[...]

PR #634, opened 2019-03-20
Grant SCL leads admin access to all SCL subproject repos

nikhita:
Ref: https://github.com/kubernetes/org/issues/630#issuecomment-474871018

@luxas was not a member of k-sigs so I've added him to the org here.

/assign @timothysc


roberthbailey:
This erroneously re-added me to a bunch of groups.

nikhita:
> This erroneously re-added me to a bunch of groups.

oh yikes, will remove.

Issue #637, opened 2019-03-20
Steering Committee has approval rights on SIG team membership


justaugustus:
### Organization or repository
kubernetes/org

### Users affected
All (@kubernetes/owners)

### Describe the issue
While looking at https://github.com/kubernetes/org/pull/636, I noticed that the PR was already approved and merged, but by a non-SIG Release Chair/Lead.

I retroactively approved the PR, to make it clear that we're fine with the change, but I'd like to discuss a little. @timothysc was able to approve as he's part of Steering, which is listed as a root approver for k/org. I already chatted with Tim offline, and he said to definitely file an issue for this.

Do we want this? 
As SIGs should be responsible for organizing themselves and Steering is really supposed to be an approver/denier of last resort, I think that we should either:
- Remove Steering as a [root approver for k/org](https://github.com/kubernetes/org/blob/master/OWNERS#L17).
- Set `no_parent_owners: true` in SIG OWNERS files

spiffxp:
I'm a pretty strong nope on `no_parent_owners` in this case, because the buck stops with the github admins here.

Removing steering from root instead

spiffxp:
FWIW: whether membership into that team should be approvable by more than sig-release is orthogonal to steering being at root, that can definitely be represented in other ways, but I consider out of scope in this issue

justaugustus:
SGTM on all points. Thanks for addressing, @spiffxp!
Closed via #638.
/close

PR #638, opened 2019-03-20
Update OWNERS and SECURITY_CONTACTS

spiffxp:
Drop steering-committee to address #637

Add myself to SECURITY_CONTACTS

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -14,4 +14,3 @@ approvers:
   - idvoretskyi
   - nikhita
   - spiffxp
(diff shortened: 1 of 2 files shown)


justaugustus:
@spiffxp -- should we have all GH owners as security contacts?

spiffxp:
>@spiffxp -- should we have all GH owners as security contacts?

The thought had occurred to me, but I think that's a longer conversation, so I opted to scope just to match subproject owners here

Issue #605, opened 2019-03-14
Descrepencies between OWNER_ALIASES of community and org repo.


imkin:
There are a lot of differences between the OWNER_ALIASES file of [community/](https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES) repo and the [org/](https://github.com/kubernetes/org/blob/master/OWNERS_ALIASES) repo

Not sure where the deviation is. But I think whenever there is a commit to either of them a job/tool should synchronize the files on either repos.

nikhita:
/assign

> But I think whenever there is a commit to either of them a job/tool should synchronize the files on either repos.

I think this is somewhat tricky because a bot would need to create a PR...and I'm not sure we want to encourage that for OWNERS files because we have a _lot_ of them across the org. :)

The solution to fixing this long-term is probably https://github.com/kubernetes/test-infra/issues/10065.

For the short-term, I'll look into fixing this by checking both the files. Going forward, we should also make sure that there are no one-off changes to `OWNERS_ALIASES` -- all changes should be a dump/copy of the `OWNERS_ALIASES` file in k/community.

imkin:
Then probably we could choose one file as a "master".
Have a presubmit job that fails when the master OWNERS_ALIASES is not same as this one.

PR #712, opened 2019-04-11
Update OWNERS_ALIASES

nikhita:
Fixes https://github.com/kubernetes/org/issues/605

This is a copy of https://github.com/kubernetes/community/blob/master/OWNERS_ALIASES.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -10,10 +10,15 @@ aliases:
     - bgrant0607
     - jdumars
     - mattfarina
+    - derekwaynecarr
+    - dims
   sig-auth-leads:
     - mikedanese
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -159,12 +152,24 @@ aliases:
...
+    - carolynvs
+    - eparis
+    - bradamant3
+  committee-product-security:
+    - philips
+    - cjcullen
+    - tallclair
+    - liggitt
+    - joelsmith
+  committee-steering:
```
nikhita: The committee aliases aren't really used anywhere in this repo so moving `committee` as a suffix is a no-op.

PR #793, opened 2019-05-06
Remove XiaoningDing from kubernetes

cblecker:
This person removed themselves from the GitHub org and we immediately re-invited them. This removes that invitation.

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -823,7 +823,6 @@ members:
 - xiang90
 - xiangpengzhao
 - xianlubird
-- XiaoningDing
 - xiaoxubeii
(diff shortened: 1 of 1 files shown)


justaugustus:
Don't they still need to be listed though?

nikhita:
> Don't they still need to be listed though?

If we remove them from the list, this will no longer try to invite them to the org. Right now, even if they want to leave the org, being on the list means they'll keep getting an invitation.

justaugustus:
Ahhh, I took the description to mean they had accidentally left the org. Don't mind me! :sweat_smile:

nikhita:
@justaugustus np :)

/hold cancel
We can always add them back if needed.

PR #804, opened 2019-05-09
Add kubernetes-feature-branch-admins group

cblecker:
Workaround until we can better figure out permissions for feature branch maintenance.


lavalamp:
Thanks. Happy to get rid of this once we figure out a process. I won't have time to figure that out any time soon, unfortunately.

PR #859, opened 2019-05-30
Add team for youtube-admins

mrbobbytables:
ref: https://github.com/kubernetes/community/issues/3727#issuecomment-497220572


nikhita:
@mrbobbytables the `youtube-admins` slack user group has some more folks...should we add them here as well?

mrbobbytables:
> @mrbobbytables the `youtube-admins` slack user group has some more folks...should we add them here as well?

🤦‍♂ Yup. Will take care of that shortly.

PR #849, opened 2019-05-29
Enforce non-admins shouldn't be team maintainers

cblecker:
This is a manual revert of https://github.com/kubernetes/org/pull/591, and a retake of https://github.com/kubernetes/org/pull/592

If this looks good, I'll do a manual peribolos run to fix any weird errors that arise.

PR #927, opened 2019-06-19
Add /me to the patch release and release managers team

hoegaarden:
- I was removed from `patch-release-team` in [this PR][pr], I am not sure why. Re-adding myself again.
- I think I also need to be on `release-managers` to be able to actually cut a patch.

If there are good reasons why I have been removed in the first place, please let me know.

/cc @kubernetes/patch-release-team 
/cc @kubernetes/release-managers 
/cc @cblecker 

[pr]: https://github.com/kubernetes/org/pull/849

### DIFF
--- a/config/kubernetes/sig-release/teams.yaml
+++ b/config/kubernetes/sig-release/teams.yaml
@@ -124,6 +124,7 @@ teams:
     members:
     - aleksandra-malinowska # 1.13 / 1.14 Patch Release Team
     - feiskyer # 1.12 Patch Release Manager
+    - hoegaarden # 1.13 / 1.14 Patch Release Team
     - tpepper # 1.13 / 1.14 Patch Release Team
     privacy: closed
   publishing-bot-reviewers:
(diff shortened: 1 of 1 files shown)


cblecker:
You were removed as only active branch managers or patch release managers should be in these teams. According to the docs in https://git.k8s.io/sig-release, I see you were the 1.14 branch manager, but that release has already shipped.

justaugustus:
@hoegaarden -- I had discussed with @cblecker (via DM) that this change should've been made separately from the changes in that PR, but should've noted that publicly as well.

Christoph --
Approving this as Hannes will be a part of the official roster moving forward: https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#contact

/lgtm
/approve

cblecker:
@justaugustus Thanks for the updated docs!

I have two asks:
- When the 1.15 dust settles, would you mind sending out a wider notice of this change? I don't remember seeing one go out to the whole project.
- Can you update the docs to note that if this team changes, you should explicitly reach out to the GitHub admin team? In addition to these teams, there's also things like k-security access that needs to be granted/updated.

justaugustus:
@cblecker -- For sure. I'm actually writing a doc on Release Managers team structure right now.
Stay tuned! :)

cblecker:
@justaugustus You're the best!

justaugustus:
@cblecker -- No, no, you!

PR #954, opened 2019-06-25
Remove rustyrobot from Kubernetes org

nikhita:
The user doesn't exist anymore, probably moved to @yoojinl.

@yoojinl can you confirm that @rustyrobot was your old username and it was renamed to @yoojinl? Will readd you after that.

This should fix peribolos.

/assign @cblecker


yoojinl:
Hi @nikhita, yes, I confirm, I've just changed my user-name.

PR #1106, opened 2019-08-15
Drop empty maintainers field

spiffxp:
It shouldn't be present if it's empty


spiffxp:
(you can file this under: yet another yaml parsing bug I ran into while trying to script a report :P)

PR #1197, opened 2019-09-18
Add chases2 to test-infra oncall

chases2:
/assign @fejta 
/cc @cjwagner

### DIFF
--- a/config/kubernetes/sig-testing/teams.yaml
+++ b/config/kubernetes/sig-testing/teams.yaml
@@ -57,12 +57,10 @@ teams:
     - fejta
     - spiffxp
     members:
-    - amwat
     - BenTheElder
+    - chases2
     - cjwagner
-    - ixdy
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -59,6 +59,7 @@ teams:
     members:
     - amwat
```
fejta: Can you drop amwat, ixdy, krzyzacy?

chases2: Should I drop them from `sig-testing-pr-reviews` and/or `test-infra-maintainers` as well?

fejta: Let's leave those alone

chases2: Changes have been made

PR #1275, opened 2019-10-08
Add hasheddan as a milestone maintainer for 1.17

hasheddan:
Request to be added as milestone maintainer for 1.17 release.

/cc @alenkacz


alenkacz:
@hasheddan is very active as a CI signal shadow, I think he definitely should be a milestone maintainer so 👍 

cc. @guineveresaenger @justaugustus 

/approve

PR #1278, opened 2019-10-08
Reuse repo-infra verify checks

fejta:
* Add verify checks from repo-infra
* Delete duplicated verify, update scripts in hack
* `bazel run @io_k8s_repo_infra//hack:update-deps` to clean up some no longer needed repos.

/assign @cblecker @nikhita


cblecker:
It looks like the verify tests are running on both the test-all and verify-all jobs?

fejta:
Yes, I'm going to delete the verify-all job as its now redundant (just like we did in test-infra), but figured I would make it run on pull-org-test-all first?

PR #1279, opened 2019-10-08
Use --config=remote in ci

fejta:
/assign @cblecker @nikhita


fejta:
woops: `ERROR: Failed to query remote execution capabilities: PERMISSION_DENIED: Request had insufficient authentication scopes.`

fejta:
Will retest after test-infra#14664 merges

PR #1359, opened 2019-10-30
Updates kompose teams

cdrage:
Updates the kompose teams outdated admin and maintainers.

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1499,31 +1499,21 @@ teams:
     description: Admin access to the kompose repo
     members:
     - cdrage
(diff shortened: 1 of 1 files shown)


mrbobbytables:
@cdrage just an FYI - the commit message can't contain github handles leading with the `@`. This is to prevent folks from unnecessary pings.

cdrage:
Removes @sarahnovotny @procrypt

Adds @hangyan

Okay! I've updated it, thanks @mrbobbytables

mrbobbytables:
Awesome, thanks for being so on the ball with this stuff 👍 
/lgtm
/approve

PR #1370, opened 2019-10-31
Update list of folks with admin access to the website repo

zacharysarah:
This PR updates the list of people with admin access to the k/website repo as part of [OWNERS cleanup](https://groups.google.com/d/msg/kubernetes-dev/4160VsBL7OI/BsBRZSqpCQAJ). 

I've pruned the list of people from SIG Docs to its current chairs:

```
    - Bradamant3
    - jimangel
    - zacharysarah
```

A couple of questions:

1. Is there a compelling reason for these folks to continue with admin permissions for k/website?

```
    - Kashomon
    - lavalamp
    - pwittrock
    - sebgoa
    - thockin
```
2. Are there folks whom we should add instead? 

I'm thinking it might be good to add one or two people from steering.

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1727,18 +1727,8 @@ teams:
     description: Admin access to the website repo
     members:
     - Bradamant3
-    - chenopis
-    - jaredbhatti
-    - Kashomon
-    - lavalamp
-    - mistyhacks
-    - pwittrock
-    - sebgoa
-    - steveperry-53
-    - tfogo
-    - thockin
+    - jimangel
     - zacharysarah
-    - zparnold
     privacy: closed
   maintainer-test-exemptions:
     description: Maintainers who for whatever random reason should not own tests.


justaugustus:
@zacharysarah -- Restricting admin access to SIG Docs Chairs seems right from my perspective.
I can't see a reason why Steering would need admin, that couldn't instead be handled by the GitHub Admin subproject.

/lgtm
/approve

For discussion:
/hold

zacharysarah:
@justaugustus 👋🏻 Thanks for the review. I’ll prune the list to current SIG Docs chairs only. 👍🏻

nikhita:
> I can't see a reason why Steering would need admin, that couldn't instead be handled by the GitHub Admin subproject.

Right. :+1: Steering doesn't need to have admin access to the repo.

nikhita:
> Is there a compelling reason for these folks to continue with admin permissions for k/website?

And this looks like just old cruft...

zacharysarah:
@justaugustus Changes made!

PR #1379, opened 2019-11-04
Add serathius to kubernetes-incubator org

serathius:
/assign @brancz 
/assign @mrbobbytables

Adding myself as owner for MetricsServer as discussed with @brancz.
@mrbobbytables I will handle migrating MetricsServer to Kubernetes SIGs

### DIFF
--- a/config/kubernetes-incubator/org.yaml
+++ b/config/kubernetes-incubator/org.yaml
@@ -451,6 +451,7 @@ teams:
     - loburm
     - ncdc
     - piosz
+    - serathius
     - sttts
     - thockin
     - x13n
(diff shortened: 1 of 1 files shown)


config/kubernetes-incubator/org.yaml
```diff
@@ -450,6 +451,7 @@ teams:
     - loburm
     - ncdc
     - piosz
+    - serathius
```
mrbobbytables: If you don't mind, please add this part as a separate commit (can still be a part of this PR, just a separate commit). The big thing we need as individual commits is being added/removed from an org. 😬

serathius: Sure.

Issue #1382, opened 2019-11-04
Migrate metrics-server to kubernetes-sigs


serathius:
### New Repo, Staging Repo, or migrate existing
Migrate existing repository from [kubernetes-incubator/metrics-server](https://github.com/kubernetes-incubator/metrics-server)

### Requested name for new repository
metrics-server

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
piosz
brancz

### If not a staging repo, who should have write access
s-urbaniak
serathius

### If not a staging repo, who should be listed as approvers in OWNERS
kawych
s-urbaniak
piosz
brancz
serathius

### If not a staging repo, who should be listed in SECURITY_CONTACTS
s-urbaniak
piosz
brancz
serathius

### What should the repo description be
by default we will set it to "created due to (link to this issue)"

### What SIG and subproject does this fall under in sigs.yaml
this is part of Metrics API subproject for sig-instrumentation

### Approvals
/cc @brancz @s-urbaniak @kawych

brancz:
Not sure where I need to approve, but this has my approval.

mrbobbytables:
That's good enough for sign off 👍

mrbobbytables:
The new teams and org memberships have been added in #1386 
The update to sigs.yaml is here: https://github.com/kubernetes/community/pull/4216

Once the teams have been merged, I will proceed with the migration of the repo 👍

mrbobbytables:
The repo has been migrated and teams have been added: https://github.com/kubernetes-sigs/metrics-server
PR is out to update OWNERS: https://github.com/kubernetes-sigs/metrics-server/pull/346

Once that PR has merged, the migration should be complete 👍

serathius:
Updated list of approvers to include all admins and writers. 
Sorry for confusion.

mrbobbytables:
All outstanding tasks have been completed. 
With that, I'm going to go ahead and close the issue 👍 
If there are any problems, please feel free to reopen or ping me on slack.

Thanks!

/close

serathius:
Thanks for your help!!

serathius:
@lilic

PR #1386, opened 2019-11-05
Add metrics server members and teams to kubernetes-sigs

mrbobbytables:
ref: https://github.com/kubernetes/org/issues/1382

Adds members and teams for metrics-server migration to kubernetes-sigs.

@s-urbaniak, @serathius, and @kawych are all members of the kubernetes org and implicitly permitted to join other orgs.

@kawych is not a part of the teams, but is being added as the [primary approver for the repo](https://github.com/kubernetes/org/issues/1382).

/area github-membership
/area github-repo

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -279,6 +279,7 @@ members:
 - runcom
 - runyontr
 - s-ito-ts
+- s-urbaniak
 - saad-ali
 - sallyom
 - SandeepPissay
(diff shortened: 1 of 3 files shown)


config/kubernetes-sigs/org.yaml
```diff
@@ -528,6 +531,18 @@ teams:
     - riverzhang
     - woopstar
     privacy: closed
+  metrics-server-admins:
```
nikhita: Can we instead create a sig-instrumentation directory and move it there?

mrbobbytables: will fix it shortly 👍

Issue #1305, opened 2019-10-14
Migrate bootkube to kubernetes-sigs


aaronlevy:
### New Repo, Staging Repo, or migrate existing
migrate existing

### Requested name for new repository

https://github.com/kubernetes-sigs/bootkube

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
@aaronlevy 

### If not a staging repo, who should have write access
@aaronlevy 

### If not a staging repo, who should be listed as approvers in OWNERS
@aaronlevy 

### If not a staging repo, who should be listed in SECURITY_CONTACTS
@aaronlevy 

### What should the repo description be
Launch a self-hosted Kubernetes cluster

### What SIG and subproject does this fall under in sigs.yaml
This is a new subproject for sig-cluster-lifecycle called bootkube

### Approvals
I believe this is already technically part of sig-cluster-lifecycle (https://github.com/kubernetes/community/tree/master/sig-cluster-lifecycle#bootkube), but the repo was never moved.

/cc @kubernetes/sig-cluster-lifecycle 

### Additional context for request
Moving from deprecated kubernetes-incubator org.

Intention is to add community (and active SIG members) to OWNERS once this is moved. /cc @andrewrynhard @rmenn

nikhita:
@aaronlevy Thanks for creating an issue to get this moving!

Regarding approval, we'll still need at least one lgtm from a SIG lead since this involves moving repos (albeit, in the kubernetes orgs).

/assign @timothysc @justinsb 
for approval

nikhita:
xref https://github.com/kubernetes/community/issues/1922

timothysc:
@aaronlevy - How active is the project now-a-days?  I was looking through the commit logs and if the most prevalent users are openshift, is there any other major consumers?

rajugupta15:
Yes, In Razorpay, We are using bootkube.

aaronlevy:
Also the Talos Systems folks have been maintaining a fork, but would like to contribute back to the upstream project. That's part of the reason for this move (can't add new people to incubator projects).

timothysc:
[...]

PR #1401, opened 2019-11-07
Add bootkube teams to kubernetes-sigs.

mrbobbytables:
ref: https://github.com/kubernetes/org/issues/1305

EDIT:Adds @aaronlevy and the bootkube teams to the kubernetes-sigs org.
aaronlevy is implicitly allowed to join k-sigs as they are a member of the kubernetes org.


mrbobbytables:
woops, looks like @aaronlevy isn't a member of k-sigs. Will add quick.

PR #1418, opened 2019-11-15
add xichengliudui to sig-docs-zh team

xichengliudui:
I hope to join the sig-docs-zh team
*  I've become reviewer/approver for zh docs.
* I worked for website repo for 15 months.
* Merged and approve pr 179 [link](https://github.com/kubernetes/website/pulls/assigned/xichengliudui)
* Personal create includes merged pr 225 [link](https://github.com/kubernetes/website/pulls/xichengliudui)

Next, I will work in zh docs for a long time, so that I can manage and approve it more conveniently! So I created this pr
/cc @zacharysarah  @chenrui333

PR #1375, opened 2019-11-02
Clarify who has website write access and why

zacharysarah:
Based on [this #sig-release Slack discussion](https://kubernetes.slack.com/archives/C2C40FMNF/p1572534834092200), this PR changes the team name of `sig-docs-l10n-admins` to `kubernetes-website-write` and adds comments for why each team member has write access.

This PR also freshens team membership to the [current SIG Docs chairs](https://github.com/kubernetes/org/blob/master/OWNERS_ALIASES#L39-L42).

Many thanks to @justaugustus for the guidance!

/sig docs

PR #1447, opened 2019-11-30
Update access to xichengliudui

xichengliudui:
#1418 describes why I have permissions, but I lost access due to the #1375 update
please let me know if I'm wrong.

### DIFF
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -299,7 +299,7 @@ teams:
     - SataQiu # L10n: Chinese
     - seokho-son # L10n: Korean
     - truongnh1992 # L10n: Vietnamese
-    - xichengliudui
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-docs/teams.yaml
```diff
@@ -277,6 +277,7 @@ teams:
     - seokho-son # L10n: Korean
     - tnir # L10n: Japanese
     - truongnh1992 # L10n: Vietnamese
+    - xichengliudui
```
zacharysarah: Please add a comment indicating that permissions are based on localization:
```suggestion
    - xichengliudui # L10n: Chinese
```

xichengliudui: I haven't been active this week, but I've been trying

jimangel:
@xichengliudui - will you please resolve the conflicting file? Otherwise
/lgtm

PR #1467, opened 2019-12-06
kubernetes-csi: add pohly as csi-test maintainer

pohly:
In practice I've been the one who has prepared all of the recent
releases, so it makes sense that I also do the last step with tagging
it.


msau42:
/lgtm
@pohly has been the primary maintainer for csi-test

PR #1481, opened 2019-12-11
add kwiesmueller to .kubernetes-maintainers.members

kwiesmueller:
As a member of wg-apply, it would be good to have the permission for editing our Umbrella Issue https://github.com/kubernetes/kubernetes/issues/73723.
Also i would like to try moving our issues into a wg-apply project.

It seems like all this is only possible with GitHub write permissions which seem to be bound to this group.

If there is another way to do this, please help.

/wg apply
/cc @apelisse

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1704,6 +1704,7 @@ teams:
     - Kashomon
     - kevin-wangzefeng
     - krousey
+    - kwiesmueller
     - lavalamp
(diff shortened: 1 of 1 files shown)


kwiesmueller:
Adding me to the group just for this might be overkill, but then there seems to be no other way to give permissions this granular on GitHub :-/

kwiesmueller:
Seems like an Org Project might do the trick.

PR #1630, opened 2020-02-12
add caesarxuchao to apiserver network proxy admin

caesarxuchao:
I need these additional permissions to cut releases, add tags, etc.

/assign @cheftako 
cc @Jefftree


Jefftree:
+1

PR #1631, opened 2020-02-12
Replace l-korbes with ellenkorbes

nikhita:
The user (https://github.com/l-korbes) is no longer present, which breaks peribolos.

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-org-peribolos/1227356355153629184

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -294,6 +294,7 @@ members:
 - ehashman
 - electrocucaracha
 - elg0nz
+- ellenkorbes
 - eloyekunle
 - ElvinEfendi
 - emaildanwilson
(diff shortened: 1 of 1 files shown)


nikhita:
This is now probably @ellenkorbes. @ellenkorbes -- can you confirm that your username has been updated to `ellenkorbes`?

ellenkorbes:
Yes, it's me, sorry, I can't make up my mind 😬😅

nikhita:
lol, np, added your new username :)

ellenkorbes:
Thanks! 😊😊

justaugustus:
I forgot who I was talking to about this, but it suggests that peribolos would be better suited retrieving users by ID, instead of username.

Yeah, yeah, I know, we're all oversubscribed :stuck_out_tongue_closed_eyes:

PR #1661, opened 2020-02-26
Add sig-storage co-chair

xing-yang:
* @jsafrane and @msau42 are sig-storage tech leads.
* @saad-ali and @xing-yang are sig-storage co-chairs.
* And remove @childsb.

See https://github.com/kubernetes/community/tree/master/sig-storage


msau42:
/assign @saad-ali 
Looks like sig-storage-leads needs to be updated here too?

PR #1667, opened 2020-02-27
Add benmoss to sig-windows-tools maintainers

benmoss:
@PatrickLang @michmike


PatrickLang:
a bit of context on this - Ben is planning to mark releases as some of the kubeadm work for alpha & beta releases are completed. That way people an download a zip/tarball instead of doing a git clone when they're trying it out

cblecker:
If you just need to cut releases, `-maintainers` is sufficient. Also, the lists need to be alpha sorted

benmoss:
Switched to just maintainers and made it sorted

justaugustus:
Can we edit the PR title to match?

PR #1671, opened 2020-02-28
sig-instrumentation: Re-add loburm, remove mxinden

brancz:
I wasn't aware loburm was on parental leave for 10w within the last half a year, so adding him back and we'll reassess in some time. Sorry!

Removing mxinden due to inactivity. Thanks Max for everything you've done and contributed!


justaugustus:
SGTM!
/lgtm
/approve
/hold -- lift after another approval from Instrumentation

PR #1675, opened 2020-02-28
Remove stale invites for kubernetes-client

### LINKED ISSUE #1627: Remove stale invites
There are a number of stale invites on all of our orgs.

These are folks that have either:
- left the org manually, and then we reinvited them
- never accepted their original invite to the org

Either way, we can safely remove these folks as these invites have been outstanding for awhile.

I opened https://github.com/kubernetes/org/pull/1626 to deal with these in @kubernetes, but we should also do the other orgs too.

Basic task:
- Look at the pending invites for an org (e.g. https://github.com/orgs/kubernetes-sigs/people)
- git blame the associated file (`git blame config/kubernetes-sigs/org.yaml | grep username`)
- If the date when they were added to the file is more than 30 days, remove them from the file
- Open 1 PR per org, just to keep things manageable for review

Complete:
- [x] kubernetes
- [ ] kubernetes-sigs
- [ ] kubernetes-incubator
- [ ] kubernetes-client
- [ ] kubernetes-csi

@justaugustus @nzoueidi @idealhack: Any of you three interested in splitting up the above orgs and taking a crack? Shouldn't take more than 20 minutes per org.

nzoueidi:
2019-11-04 19:00:11 -0500  37) - guoshimin

This should be the last one to close #1627

### DIFF
--- a/config/kubernetes-client/org.yaml
+++ b/config/kubernetes-client/org.yaml
@@ -34,7 +34,6 @@ members:
 - fabianvf
 - flavio
 - grodrigues3
-- guoshimin
 - iamneha
 - idealhack
 - irvifa


nzoueidi:
cc @cblecker

PR #1738, opened 2020-03-22
Update Italian localization teams & corrected language affilation

rlenferink:
- Synced Italian teams with k8s/website repo (as soon as https://github.com/kubernetes/website/pull/19790 merges)
- Corrected localization affiliation from Italian -> German (as German approver)

### DIFF
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -311,7 +311,7 @@ teams:
     - potapy4 # L10n: Russian
     - raelga # L10n: Spanish
     - remyleone # L10n: French
-    - rlenferink # L10n: Italian
+    - rlenferink # L10n: German
     - SataQiu # L10n: Chinese
     - seokho-son # L10n: Korean
     - truongnh1992 # L10n: Vietnamese
--- a/config/kubernetes/sig-docs/teams.yaml
+++ b/config/kubernetes/sig-docs/teams.yaml
@@ -132,14 +132,13 @@ teams:
     members:
     - fabriziopandini
     - mattiaperi
-    - rlenferink
+    - micheleberardi
   sig-docs-it-reviews:
     description: PR reviews for Italian content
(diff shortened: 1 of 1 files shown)


rlenferink:
/cc @sftim
Now that the website PR is merged this can continue as well. Thanks already!

config/kubernetes/sig-docs/teams.yaml
```diff
@@ -311,7 +311,7 @@ teams:
     - potapy4 # L10n: Russian
     - raelga # L10n: Spanish
     - remyleone # L10n: French
-    - rlenferink # L10n: Italian
+    - rlenferink # L10n: German
```
kbhawkey: Note: This file seems outdated.

rlenferink: At least for me it is, therefore the update. As far as I know we try to have someone from each localization to be part of this team, however IMO we also need to be careful since it grants write access to the website repo.

rlenferink:
@sftim because this syncs the team with the k8s/website repo (at least the Italian ones), can you `/approve` this / point out whether changes need to be made before merging this?

sftim:
Not sure I should /approve and /lgtm the same PR; can someone guide me on that?

config/kubernetes/sig-docs/teams.yaml
kbhawkey: OK, @rlenferink ,
Pinging the Docs chair and team leads,
@zacharysarah , @kbarnard10 @jimangel

rlenferink:
> Not sure I should /approve and /lgtm the same PR; can someone guide me on that?

Sure! `/lgtm` and `/approve` in the same PR happens more. See #1748 as example. Depends on the change. In the case of this particular PR I'd be fine if someone LGTM-ed and approved. If it's about creating a new team / granting someone write/admin access I would want to have different reviewers/approvers.

PR #1909, opened 2020-06-01
Add kbruner as milestone maintainer

kbruner:
/cc @gianarb @droslean


gianarb:
Thanks! 

I think this is ready now

/assign @justaugustus

PR #1919, opened 2020-06-06
sig-release: Cleanup teams

justaugustus:
We're discussing consolidation of the Patch Release Team and Branch Managers into a single group, "Release Managers" in https://github.com/kubernetes/sig-release/pull/1106.

While I was doing that, I decided to also do a little cleanup.

This PR does the following:
- Remove `patch-release-team` notification team
- Nest subproject teams
- Prune `sig-release` GitHub team
- Lint team descriptions to 80-char length

Before https://github.com/kubernetes/sig-release/pull/1106 or this can merge, I need to prep PRs to change any references from @kubernetes/patch-release-team to @kubernetes/release-managers.
ref: https://cs.k8s.io/?q=patch-release-team&i=nope&files=&repos=
/hold

PTAL and let me know if anything looks amiss.

/assign @tpepper 
cc: @kubernetes/release-engineering @kubernetes/release-team @kubernetes/licensing @kubernetes/sig-release


config/kubernetes/sig-release/teams.yaml
```diff
@@ -143,103 +143,26 @@ teams:
     - nikhita
     - spiffxp
     members:
```
tpepper, line 146: What's the pruning criteria applied here?  I'm surprised by some of the names being dropped versus no additions either.

config/kubernetes/sig-release/teams.yaml
```diff
@@ -58,7 +27,7 @@ teams:
     - cheftako # Cloud Provider
     - childsb # Storage
     - chrigl # OpenStack
-    - cpanato # Branch Manager
```
tpepper, line 61: These hunks maybe got a 'git add -i' in the wrong commit?  Maybe just a nit though if the commits are to be squashed?

config/kubernetes/sig-release/teams.yaml
tpepper: From discussion with @justaugustus : 

This list is the explicitly named individuals, plus the nested groups.  Anybody in a nested group doesn't need to be in the top level list, as they are already implicitly include.  Anybody who's not been recently active in the SIG gets removed periodically.  Anybody who's active in the SIG (and not a lead or other role in a nested list) will be added periodically.

tpepper:
/lgtm
after discussion with @justaugustus

justaugustus:
Thanks for the reviews everyone!
/hold cancel

PR #1920, opened 2020-06-09
update sig testing

BenTheElder:
/cc @spiffxp @nikhita


BenTheElder:
fixed (I think? too bad it's not automated)

PR #1978, opened 2020-07-01
Add wongma7 to ebs, efs, and fsx maintainers

wongma7:
Adding self to ebs, efs, and fsx repos.

Unsure if this is sufficient for permissions to write releases but that is the goal here. Otherwise there is only 1 active maintainer @leakingtapan making releases.

/cc @jsafrane 
/cc @leakingtapan


nikhita:
Approving as per @jsafrane's lgtm since they are an admin on all of these teams

/approve

PR #1994, opened 2020-07-06
Remove karan from sig-cluster-lifecycle teams

karan:
:/


neolit123:
@karan i'm going to send a PR clearing the teams and making them more up to date.

PR #2044, opened 2020-07-17
Renaming secret-sync-rotation to k8s-gsm-tools

spiffxp:
ref: https://github.com/kubernetes-sigs/secret-sync-rotation/issues/3


spiffxp:
Thanks for the catch, updated

b01901143:
Thanks.
/lgtm

PR #2047, opened 2020-07-18
Remove inactive members

mrbobbytables:
ref: https://github.com/kubernetes/org/issues/2013
/assign @cblecker @nikhita 

FYI regarding team removal: Members that were a part of a team should be removed, but if the team is still in the main org.yaml file, it won't have the additional commit msg.

I'll upload the script I used to do it in a separate PR. 

/hold

PR #2048, opened 2020-07-18
[WIP] Add remove-members helper script

mrbobbytables:
This is the script used to generate #2047 
It needs some final polish like usage information, but figured I'd share for initial review / vetting alongside #2047

/hold

### DIFF
--- a/hack/remove-members.sh
+++ b/hack/remove-members.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly REPO_ROOT="$(git rev-parse --show-toplevel)"
+readonly CONFIG_PATH="$REPO_ROOT/config"
+readonly DRYRUN="${DRYRUN:-true}"
+
+members=()
+mapfile -t members < "$1"
+for member in "${members[@]}"; do
+  matches=()
+  orgs=()
+  teams=()
+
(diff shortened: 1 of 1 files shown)


hack/remove-members.sh
```diff
@@ -0,0 +1,90 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly REPO_ROOT="$(git rev-parse --show-toplevel)"
+readonly CONFIG_PATH="$REPO_ROOT/config"
+readonly DRYRUN="${DRYRUN:-true}"
```
nikhita, line 22: Can we make this a flag?

hack/remove-members.sh
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly REPO_ROOT="$(git rev-parse --show-toplevel)"
+readonly CONFIG_PATH="$REPO_ROOT/config"
```
nikhita, line 21: Can/should we make this configurable? In cases where someone is active in `kubernetes-sigs` but not active in `kubernetes`, or if we just want to remove members from `kubernetes`?

hack/remove-members.sh
mrbobbytables: Yeah, this was me being kind of lazy to get it done lol

hack/remove-members.sh
mrbobbytables: I can do that 👍

mrbobbytables:
There is one issue I just thought of, it currently won't remove members if they have a comment after their name like
` - foo # herp derp`
Lucked out with the original batch

fejta:
Thoughts on putting this somewhere aside from hack? The unfortunately named hack folder generally seems to be used for lint-type updates and/or verifications. 

Isn't this an operational thing? AKA I want to remove Erick from the org and this is the tool to do so, which just happens to be in bash right now.

justaugustus:
Adding on to @fejta's, I think this is important enough and of sufficient complexity that it should be written in Go.

mrbobbytables:
I'm okay with that -- I'm a little low on time at the moment, but I'll see what I can do to get something a bit better pushed up. Either way will get something before the next round of clean up. 👍 

/close

nikhita:
Created https://github.com/kubernetes/org/issues/2052 so that we don't forget this. I'm also happy to take a look at it later :)

PR #2060, opened 2020-07-24
add current actual google build admins to build-admins

BenTheElder:
unclear which of the other entries should be removed

See: https://github.com/kubernetes/sig-release/issues/1150

### DIFF
--- a/config/kubernetes/sig-release/teams.yaml
+++ b/config/kubernetes/sig-release/teams.yaml
@@ -208,15 +208,15 @@ teams:
               the Kubernetes project. This team is a notification-only group,
               which includes the SIG Release Chairs for continuity.
             members:
+            - aleksandra-malinowska
             - amwat
             - BenTheElder
-            - MushuEE
-            - spiffxp
-            - aleksandra-malinowska
             - calebamiles
             - justaugustus
             - listx
+            - MushuEE
             - ps882
+            - spiffxp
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-release/teams.yaml
```diff
@@ -208,6 +208,10 @@ teams:
               the Kubernetes project. This team is a notification-only group,
               which includes the SIG Release Chairs for continuity.
             members:
+            - amwat
```
MushuEE, line 211: Also maybe alpha? Unless we are taking over the whole block

spiffxp: I think CI might fail on non-alpha anyway

BenTheElder: I think maybe the others should be removed but not clear.
humans alpha sorting things by hand is just the worst.

BenTheElder:
>   config_test.go:251: The team build-admins in org Kubernetes has org admins listed as members; these users should be in the maintainers list instead, and cannot be on the members list: spiffxp

... really? is this a github limitation or self-imposed?

>    config_test.go:251: The team build-admins in org Kubernetes has an unsorted list of members

fun

spiffxp:
>... really? is this a github limitation or self-imposed?

github

PR #2085, opened 2020-08-06
Update teams.yaml

e-blackwelder:
Adding self to test-infra-admins for oncall responsibilities. (But will it work when I'm not already a member of the community??)


BenTheElder:
> Adding self to test-infra-admins for oncall responsibilities. (But will it work when I'm not already a member of the community??)

you should file to join the project first

e-blackwelder:
Fixed in #2145!

PR #2143, opened 2020-08-27
remove mcs-api-admins team from sig service catalog teams

jberkhahn:
This team was added to SIG Service Catalog in error. I think it's supposed to be part of SIG multicluster, but I'm not really sure. I can edit the PR to add it to there if that is in fact where it's supposed to be.

/cc @pmorie 
/cc @JeremyOT

### DIFF
--- a/config/kubernetes-sigs/sig-service-catalog/teams.yaml
+++ b/config/kubernetes-sigs/sig-service-catalog/teams.yaml
@@ -1,10 +1,4 @@
 teams:
-  mcs-api-admins:
-    description: Admin access to the mcs-api repo
-    members:
-    - JeremyOT
-    - pmorie
-    privacy: closed
   service-catalog-admins:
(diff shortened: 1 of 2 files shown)


config/kubernetes-sigs/sig-service-catalog/teams.yaml
```diff
@@ -1,10 +1,4 @@
 teams:
-  mcs-api-admins:
-    description: Admin access to the mcs-api repo
-    members:
-    - JeremyOT
-    - pmorie
-    privacy: closed
```
spiffxp, line 7: Good catch! Can you move this to config/kuberntes-sigs/sig-multicluster/teams.yaml?

jberkhahn:
@spiffxp updated

nikhita (approved):
/lgtm
/approve

Issue #2144, opened 2020-08-27
REQUEST: New membership for e-blackwelder


e-blackwelder:
### GitHub Username
@e-blackwelder

### Organization you are requesting membership in
@kubernetes

### Requirements
- [+] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [+] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [+] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [+] I am actively contributing to 1 or more Kubernetes subprojects
- [+] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [+] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- @spiffxp
- @hasheddan

### List of contributions to the Kubernetes project
- SIG Testing
- https://github.com/kubernetes/test-infra/pull/18001
- https://github.com/kubernetes/test-infra/pull/18693
- https://github.com/kubernetes/test-infra/pull/17244
- Prow oncall rotation

hasheddan:
+1 from me! :tada:

spiffxp:
+1 from sponsor

spiffxp:
Hi @e-blackwelder, thanks for all your contributions so far!

I've created PR #2145 to add you to the Kubernetes Org. Once that gets merged, you should get a membership invite notification.

Welcome to @kubernetes! 🎉

/meowvie

PR #2145, opened 2020-08-27
Add e-blackwelder to kubernetes org

spiffxp:
Fixes https://github.com/kubernetes/org/issues/2144

(Didn't notice any other eligible requests while I was doing this)


spiffxp:
```
    config_test.go:246: members for Kubernetes org are unsorted
```

Welp.  I had a 50/50 chance at correcly guessing how "e-" was going to alphasort while I waited for bazel to tell me.  Looks like I lost.

PR #2268, opened 2020-10-15
Add jsturtevant to sig-windows-tools-maintainers

jsturtevant:
Helping out @benmoss with maintenance and releases.

cc: @marosset @michmike


michmike (approved):
/approve
/lgtm

PR #2277, opened 2020-10-20
add gujingit to sig-cloud-provider-alibaba-admins

gujingit:
update sig-cloud-provider-alibaba-admins


cheftako:
/approve
If @aoxn is ok with it, approving.

Issue #2348, opened 2020-11-28
REQUEST: New membership for reylejano


reylejano:
### GitHub Username
@reylejano

### Organization you are requesting membership in
@kubernetes 

### Requirements
- [X] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [X] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [X] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [X] I am actively contributing to 1 or more Kubernetes subprojects
- [X] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [X] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- @annajung 
- @jeremyrickard 
- @kikisdeliveryservice 

### List of contributions to the Kubernetes project
https://github.com/kubernetes/website/pull/25236
https://github.com/kubernetes/website/pull/24891
https://github.com/kubernetes/website/pull/23306

annajung:
+1 

@reylejano has been part of sig release for multiple releases and currently is a doc shadow for 1.20 release.

nikhita:
@reylejano looks like you were already part of the @kubernetes org but the bot removed you because of the change in username (`s/reylejano-rxm/reylejano`). The bot can't auto-update membership if username changes.

To get re-added to the org, you don't need new sponsors. :)

I have created https://github.com/kubernetes/org/pull/2349 to get you re-added to the org. Let me know if you need any more help on this! :smile:

justaugustus:
Thanks Nikhita! Beat me to it :)

reylejano:
@nikhita and @justaugustus Thank you!

justaugustus:
Welcome back! :)

Issue #1555, opened 2020-01-13
REQUEST: New membership for reylejano-rxm


reylejano-rxm:
### GitHub Username
@reylejano-rxm

### Organization you are requesting membership in
@kubernetes

### Requirements
- [x] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [x] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [x] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [] I am actively contributing to 1 or more Kubernetes subprojects
- [x] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [x] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- @evillgenius75 
- @alejandrox1
- @lachie83 

### List of contributions to the Kubernetes project
- 1.18 Release Notes Shadow

alejandrox1:
+1
Welcome to the team :)

evillgenius75:
+1
Let's have some fun!

alejandrox1:
/assign @nzoueidi 
ptal

reylejano-rxm:
Updated request to add @lachie83 as a sponsor.

nzoueidi:
My bad @reylejano-rxm, @evillgenius75 is already in OWNER file of sig release.

nzoueidi:
Hey there @reylejano-rxm, thank you for all your contributions so far :)

I've created PR #1570 to add you to the Kubernetes Org. Once that gets merged, you should get a membership invite notification.

Welcome to @kubernetes! 🎉

PR #2349, opened 2020-11-28
Fix reylejano's username

nikhita:
Fixes https://github.com/kubernetes/org/issues/2348

https://github.com/reylejano-rxm does not exist anymore and they have
changed their username to https://github.com/reylejano.

Ref:
- https://github.com/kubernetes/org/issues/1555
- https://github.com/kubernetes/org/issues/2348

This fixes peribolos - https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-org-peribolos/1332512581797023744

/assign @palnabarun 
fyi @reylejano

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -854,7 +854,7 @@ members:
 - RenaudWasTaken
 - rendhalver
 - resouer
-- reylejano-rxm
+- reylejano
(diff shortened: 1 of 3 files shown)


justaugustus:
/lgtm
/approve

+1 to re-membership as a SIG Release approver

Issue #2350, opened 2020-12-01
Migrate https://github.com/liggitt/kubectl-check-ownerreferences to kubernetes-sigs


liggitt:
### New Repo, Staging Repo, or migrate existing
migrate https://github.com/liggitt/kubectl-check-ownerreferences/ to kubernetes-sigs

### Requested name for new repository
kubectl-check-ownerreferences

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
deads2k, liggitt, caesarxuchao

### If not a staging repo, who should have write access
deads2k, liggitt, caesarxuchao

### If not a staging repo, who should be listed as approvers in OWNERS
deads2k, liggitt, caesarxuchao

### If not a staging repo, who should be listed in SECURITY_CONTACTS
deads2k, liggitt, caesarxuchao

### What should the repo description be
Read-only tool to check metadata.ownerReferences

### What SIG and subproject does this fall under in sigs.yaml
This is part of the control-plane-features subproject of sig-apimachinery (https://github.com/kubernetes/community/tree/master/sig-api-machinery#control-plane-features)

### Approvals

Proposed and approved by sig-apimachinery TLs @deads2k and @lavalamp 

### Additional context for request

Related to tooling to diagnose issues like http://issue.k8s.io/65200 and http://issue.k8s.io/92743

liggitt:
/assign @deads2k @lavalamp 
for ack

nikhita:
/assign

@liggitt can you move the repo to https://github.com/kubernetes-purgatory? :) I can then transfer the repo to k-sigs.

Also, can you also add the [boilerplate header](https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.sh.txt) to the bash scripts when you get a chance? After the transfer is also fine.

liggitt:
> @liggitt can you move the repo to https://github.com/kubernetes-purgatory? :) I can then transfer the repo to k-sigs.
> 
> Also, can you also add the [boilerplate header](https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.sh.txt) to the bash scripts when you get a chance? After the transfer is also fine.

both done 

https://github.com/kubernetes-purgatory/kubectl-check-ownerreferences

nikhita:
Repo has been migrated - https://github.com/kubernetes-sigs/kubectl-check-ownerreferences

Also created:
[...]

PR #2351, opened 2020-12-01
Add GitHub team for kubectl-check-ownerreferences

nikhita:
Ref: https://github.com/kubernetes/org/issues/2350

Adding just the admins team because the list of folks with admin and write access is the same.

/assign @mrbobbytables


nikhita:
bump

(looks like prow didn't pick up the lgtm label)

liggitt:
did this get propagated? I don't seem to have write access to the repo to configure required CI jobs

mrbobbytables:
It should be set now 👍

PR #2357, opened 2020-12-03
Refine too many args for fmt string

hi-rustin:
Refine too many args for fmt string.

Same time I think the String maybe wrong.
Like:
key = b, value = "b.yaml"
key = a, value = ""

It will generate `b=b.yaml,a,a=`.

### DIFF
--- a/cmd/merge/main.go
+++ b/cmd/merge/main.go
@@ -45,6 +45,7 @@ func (fm flagMap) String() string {
 	for key, value := range fm {
 		if value == "" {
 			parts = append(parts, key)
+			continue
 		}
 		parts = append(parts, key+"="+value)
 	}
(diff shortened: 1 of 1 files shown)


cmd/merge/main.go
```diff
@@ -96,11 +97,11 @@ func main() {
 func unmarshal(path string) (*org.Config, error) {
```
fejta, line 97: Any interest in checking the returned error on L93?

cmd/merge/main.go
```diff
@@ -45,8 +45,9 @@ func (fm flagMap) String() string {
 	for key, value := range fm {
 		if value == "" {
 			parts = append(parts, key)
+		} else {
```
fejta, line 48: nit: I'd probably write this as

```golang
if value == "" {
  parts = foo()
  continue
}
parts = foo()
```

(to always try and keep indentation minimized)

fejta (approved):
/hold cancel

fejta:
/hold
I meant

fejta (approved):
/hold cancel

nikhita:
/lgtm

Thanks @hi-rustin! :smile:

PR #2469, opened 2021-01-29
Retire from my role as Release Associate

gianarb:
Hey! After a few iterations in sig-release and after a few months helping
as an Associate, I am stepping back.

I recently spoke with a couple of teammates in the sig release and I value
the title I am holding, and I am sure other people from the community
want to contribute as an Associate. Now I have a better idea about how to
contribute and I don't need a title for that.

### DIFF
--- a/config/kubernetes-sigs/sig-release/teams.yaml
+++ b/config/kubernetes-sigs/sig-release/teams.yaml
@@ -65,7 +65,6 @@ teams:
     - ameukam # Release Manager Associate
     - cpanato # Release Manager
     - feiskyer # Release Manager
-    - gianarb # Release Manager Associate
     - hasheddan # subproject owner / Release Manager
     - idealhack # Release Manager
     - jimangel # Release Manager Associate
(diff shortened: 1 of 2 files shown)


config/kubernetes/org.yaml
```diff
@@ -362,7 +362,6 @@ members:
 - geojaz
 - gerred
 - ggriffiths
-- gianarb
```
justaugustus, line 365: You don't want to remove yourself here, as this will remove you from the Kubernetes GitHub org altogether.
...unless you also want to do that? 😿

justaugustus (changes requested):
@gianarb -- Hope to still see you around occasionally! Just one question on the review.
/hold

config/kubernetes/org.yaml
gianarb: I am not sure about how it works, technically I don't have any reason to stay in right now, I a not a member of any other team :) so, I am happy to be part of it if it does not cause any issue ❤️

justaugustus: (Resolved via DM.)

justaugustus:
Thanks and take care for now, Gianluca! 💕 

/lgtm
/approve
/hold cancel

PR #2473, opened 2021-01-29
Setup SIG Testing subteams

spiffxp:
specifically:

- add sig-testing/sig-testing-leads
- move sig-testing-pr-reviews under sig-testing

---

META: this was mostly a PR to verify what creating a new PR from an old local clone looks like following a rename of this repo's default branch

### DIFF
--- a/config/kubernetes/sig-testing/teams.yaml
+++ b/config/kubernetes/sig-testing/teams.yaml
@@ -28,26 +28,35 @@ teams:
     - timothysc
     - xiangpengzhao
     privacy: closed
-  sig-testing-pr-reviews:
-    description: sig-testing PR reviews
-    maintainers:
-    - fejta
-    - spiffxp
-    members:
-    - akutz
-    - amwat
-    - BenTheElder
-    - chases2
-    - cjwagner
-    - clarketm
-    - ixdy
-    - Katharine
-    - krzyzacy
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -28,26 +28,35 @@ teams:
...
+        - spiffxp
+        members:
+        - akutz
+        - amwat
+        - BenTheElder
+        - chases2
+        - cjwagner
+        - clarketm
+        - ixdy
+        - Katharine
```
BenTheElder, line 53: this list could probably do with some tidying?

spiffxp: agreed, I wanted to do that as a followup

spiffxp:
milestone_applier failed to apply the v1.21 milestone for the non-existant `master` branch, which is what I expected, but I don't see any errors/logs about this in k8s-prow's logs

spiffxp:
probably because it exits early/silently if `main` isn't configured, ref: https://github.com/kubernetes/test-infra/blob/4ddcfd375fcd021a16f696aa83b4bbbbd5c9ba91/prow/plugins/milestoneapplier/milestoneapplier.go#L84-L88

PR #2532, opened 2021-03-01
Update membership to reflect new username.

rohitagarwal003:


nikhita (approved):
Thanks for updating your username here. Most folks forget to do that when they change their username :)

xref https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-org-peribolos/1366234709549387776

/lgtm
/approve

config/kubernetes/org.yaml
```diff
@@ -877,6 +876,7 @@ members:
 - robinpercy
 - robscott
 - rojkov
+- rohitagarwal003
```
nikhita, line 879: oops, this needs to be sorted alphabetically :)

rohitagarwal003: Done.

PR #2575, opened 2021-03-15
Add remove-members helper tool implemented in go

### LINKED ISSUE #2052: Create tool for removing org members
We intend to start doing a regular cleanup of inactive members (https://github.com/kubernetes/org/issues/2013, https://github.com/kubernetes/org/pull/2047).

Removing org members means removing them from the config files in this repo. Removing them manually is tedious so we need to create an automated tool that can help us do that.

https://github.com/kubernetes/org/pull/2048 has prior art in bash, we probably want the new tool to be written in Go.

daemon1024:
Fixes #2052 

This tool is implementation of the bash script in go with a few changes like using flags instead of environment variables

This currently resides in the hacks folder which I believe is not the right place, let me know where I should move this to and if it deserves it's own package


hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var dryrun bool
+var repoRoot string
+
+func main() {
+	flag.StringVar(&repoRoot, "root", ".", "Root of the repository")
```
nikhita, line 21: Instead of allowing to specify the repoRoot, we could probably take the config as a flag. For example, if we want to selectively remove only members in `kubernetes`, we should be able to specify that

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+	flag.BoolVar(&dryrun, "dryrun", true, "Enable Dryrun or not")
```
nikhita, line 22: For the usage - can you elaborate on what enabling/disabling dryrun would do?

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+
+	flag.Parse()
+
+	if len(flag.Args()) < 1 {
+		fmt.Print("Usage: remove-members [flags] <file-containing-members-list>\n\nFlags:\n\t-root\tstring\tpath to root directory of the repository\n\t-dryrun\tboolean\tEnable/Disable dryrun (Default: Enabled)")
```
nikhita, line 27: Would be better to use `flag.Usage` - https://golang.org/pkg/flag/#pkg-variables

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+var dryrun bool
+var repoRoot string
+
+func main() {
+	flag.StringVar(&repoRoot, "root", ".", "Root of the repository")
+	flag.BoolVar(&dryrun, "dryrun", true, "Enable Dryrun or not")
+
+	flag.Parse()
+
+	if len(flag.Args()) < 1 {
```
nikhita, line 26: also maybe validate that only one argument is specified

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+	err = removeMembers(memberList, configPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+}
+
+//readMemberList reads the list of members to be removed from the given filepath
+func readMemberList(path string) ([]string, error) {
+	file, err := os.Open(path)
```
nikhita, line 47: @mrbobbytables is the inactive member list a yaml file? how is that generated?

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+		if dryrun == true {
+			return true, nil
+		}
+
+		updatedContent := re.ReplaceAll(content, []byte(""))
+		err = ioutil.WriteFile(path, updatedContent, 0666)
+
+		if err != nil {
+			return false, err
+		}
```
nikhita, line 136: can combine the if condition on line 132

(same changes at other places in this file)

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return members, scanner.Err()
+}
+
+//removeMembers walks through the config directory and removes the occurences of the given member name
+func removeMembers(memberList []string, configPath string) error {
+	for _, member := range memberList {
+		var orgs, teams []string
+		count := 0
+		fmt.Print(member)
```
nikhita, line 66: would be better to `log` here and a more descriptive log message

hack/remove-members.go
```diff
@@ -0,0 +1,184 @@
...
+func removeMemberFromFile(member string, path string) (bool, error) {
+
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return false, err
+	}
+
+	re := regexp.MustCompile(`(\s+)?- ` + member + `(.*)?`)
+
+	if re.Match(content) {
```
nikhita, line 124: we can probably count how many times `member` occurs in the `path` and return the count. This will ensure that `count` is accurate.

nikhita (commented):
@daemon1024 thanks for creating the PR for this!

Left some suggestions, but looks great overall! :tada:


Can you also add a test for this?

hack/remove-members.go
daemon1024: Should I change all instances of `fmt` to `log`?

Currently the output looks like 
```
~/P/o/hack (remove-members-tool|…) ❯❯❯  go run remove-members.go -path=../config/kubernetes members.txt
nikhita
 Orgs: [kubernetes]
 Teams: [sig-contributor-experience sig-release]
 Number of occurences: 3

Commit Command: "commit -m Remove nikhita from the kubernetes org -m Remove nikhita from sig-contributor-experience, sig-release teams"

daemon1024
 Orgs: []
 Teams: []
 Number of occurences: 0
```

P.s. I don't really want to remove you :P just for showing the output :)

hack/remove-members.go
daemon1024: Do you mean to count the number of matches in the said file?

hack/remove-members.go
daemon1024: Updated Usage
```
Usage: remove-members [flags] <file-containing-members-list>
  -dryrun
        Enable Dryrun or not. This flag controls whether the changes are simulated or live. If the changes are simulated it will only print the removal details and the commit message (default true)
  -path string
        Path to config directory/subdirectory (default ".")
```

daemon1024:
> Can you also add a test for this?

Would it be fine to add some test `.yaml` files cause it wouldn't be right to perform tests on the actual files. Or should we work with [TempFile](https://pkg.go.dev/io/ioutil#TempFile)

hack/remove-members.go
```diff
@@ -0,0 +1,179 @@
...
+	"sort"
+	"strings"
+)
+
+var dryrun bool
+var configPath string
+
+func main() {
+	flag.StringVar(&configPath, "path", ".", "Path to config directory/subdirectory")
+	flag.BoolVar(&dryrun, "dryrun", true, "Enable Dryrun or not. This flag controls whether the changes are simulated or live. If the changes are simulated it will only print the removal details and the commit message")
```
nikhita, line 22: @mrbobbytables for feedback on the help text

hack/remove-members.go
```diff
@@ -0,0 +1,179 @@
...
+
+	flag.Parse()
+
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage: remove-members [flags] <file-containing-members-list>\n")
```
nikhita, line 27: would be good to list out the flags here. http://courses.cms.caltech.edu/cs11/material/general/usage.html has some details on the format

hack/remove-members.go
nikhita: I think fmt is also ok, but would be good to use a more descriptive message :)

hack/remove-members.go
nikhita: > Do you mean to count the number of matches in the said file?

Yeah :)

nikhita:
> Would it be fine to add some test .yaml files cause it wouldn't be right to perform tests on the actual files. Or should we work with TempFile

It would be fine to add test yaml files :+1:

hack/remove-members.go
daemon1024: `flag.PrintDefaults()` prints the details of the flag. The current output is 
```
Usage: remove-members [flags] <file-containing-members-list>
  -dryrun
        Enable Dryrun or not. This flag controls whether the changes are simulated or live. If the changes are simulated it will only print the removal details and the commit message (default true)
  -path string
        Path to config directory/subdirectory (default ".")
```

Or do you mean to provide the flags in the header line itself?

daemon1024: `Usage: remove-members [-dryrun] [-path] member-file (file-containing-members-list)`

Is this fine?

hack/remove-members.go
mrbobbytables: It's been a text file with 1 member per line. The list is generated from this sheet after some manual vetting: https://docs.google.com/spreadsheets/d/1jqxMOo9f1EG72i4sGE7g6RlNPvQ4_qfOx9K-UM4YsLw/edit

nikhita: Thanks! One more follow up question - how does one generate this sheet?

Want to remove you as a single point of failure :)

mrbobbytables: The big stuff to regen it are in the notes tab in the sheet 👍 I thought it was owned/under the sig contribex shared dir, but it looks like I'm the only owner atm. I'll move it over in a bit.

hack/remove-members.go
mrbobbytables: Not a blocker for this, but potentially something in the future. We can marshal/unmarshal yaml while leaving the comments (https://github.com/go-yaml/yaml/issues/132). This would also let us get team names directly.

hack/remove-members.go
mrbobbytables: It could probably be simplified down to  something like: `Simulates changes to be applied and prints removal details`

hack/remove-members.go
```diff
@@ -0,0 +1,179 @@
+package main
+
+import (
+	"bufio"
+	"flag"
```
palnabarun, line 5: Can we use https://github.com/spf13/pflag instead of flag? This will enable us to have POSIX style flags. (`--foo`)

hack/remove-members.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	defer file.Close()
+
+	var members []string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		members = append(members, scanner.Text())
+	}
+	return members, scanner.Err()
+}
```
palnabarun, line 61: How about simplifying this using `ioutil.ReadFile` as used in `removeMemberFromFile`?

Note: `ioutil.ReadFile` has been deprecated in Go 1.16 although it shouldn't be a blocker since this repository still uses Go 1.13.

hack/remove-members.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if re.Match(content) {
+
+		//Mofify the file only if it's not a dry run
+		if dryrun == true {
+			return true, nil
+		}
+
+		updatedContent := re.ReplaceAll(content, []byte(""))
+		if err = ioutil.WriteFile(path, updatedContent, 0666); err != nil {
```
palnabarun, line 133: We should be preserving the mode of the file that we just read instead of writing it with `0666`. Can you please fix that?

palnabarun (changes requested):

hack/remove-members.go
palnabarun: @mrbobbytables -- I had a dig at this earlier by parsing the org files using [k8s.io/test-infra/prow/config/org.Config](https://github.com/kubernetes/test-infra/blob/master/prow/config/org/org.go), but there seemed to be something off there.

The processed org config when written back to disk had a lot of diff compared to the original file.

We can possibly dig deeper into it later.

daemon1024:
Should I move this to a new directory? ( `helpers/remove-members` maybe?

Since I added an external package it seems to adding it to the go.mod at the root, I haven't committed it yet cause I am not sure if this a part of that or should I initiate a seperate go.mod for this.

palnabarun:
@daemon1024 

> Should I move this to a new directory? ( helpers/remove-members maybe? )

You can move this to `cmd/remove-members/main.go`; that sounds to me as the most logical place.

> Since I added an external package it seems to adding it to the go.mod at the root, I haven't committed it yet cause I am not sure if this a part of that or should I initiate a separate go.mod for this.

I think it should be fine to have the `pflag` dependency in root go.mod.

hack/remove-members.go
mrbobbytables: It's been a bit since I looked at it , but potentially whitespace =/ since yaml supports some fuzzy inclusion:
```
foo:
  - bar
```
is the same as:
```
foo:
- bar
```

Theres what people have written (both formats) and what go will want to format it as.

palnabarun: That as well. we may want to bite the bullet with a big initial diff as the subsequent updates _should_ be deterministic .

One other thing was, in order to process comments, the structs in `k8s.io/test-infra/prow/config/org.Config` should have fields with `yaml.Node` as field type and then have post-processing helper method. I had taken a step back because of that complexity.

mrbobbytables: Yeah, what I'd probably do at that point is get away from editing by hand, and then have an add-member function or have a make command that ensures proper formatting before commit. Out of scope for this one though^^;;;

daemon1024:
We need to print out the removal details, what would be the preferred method from the following:-
- Executing `git diff`
- Running for loops and finding differences between the content
- Use some package like https://github.com/sergi/go-diff

palnabarun:
What do you want to exactly print? If you want the diff between the state before and after executing this script, a `git diff` or `git log -p commit1..commit2` should suffice.

daemon1024:
> What do you want to exactly print? 

When the program is in dry run, I wanted to print the details of what will be removed ( "simulation" ? )

cmd/remove-members/main_test.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestReadMemberList(t *testing.T) {
+	memberList, err := readMemberList("test/members.txt")
```
spiffxp, line 9: Would prefer you move this to `testdata`, it's more idiomatic (see `go help test`)
```suggestion
	memberList, err := readMemberList("testdata/members.txt")
```

cmd/remove-members/test/org.yaml
```diff
@@ -0,0 +1,3 @@
+members:
```
spiffxp, line 1: another test case to consider: an org.yaml that also has a `teams` entry

cmd/remove-members/test/teams.yaml
```diff
@@ -0,0 +1,15 @@
+teams:
```
spiffxp, line 1: another test case to consider: a teams.yaml in a subpath

cmd/remove-members/main_test.go
```diff
@@ -0,0 +1,28 @@
...
+	if !reflect.DeepEqual(memberList, expectedMemberList) {
+		t.Errorf("Values mismatch. Expected: %v Actual: %v", expectedMemberList, memberList)
+	}
+}
+
+func TestRemoveMembers(t *testing.T) {
+
+	memberList := []string{"johndoe", "janedoe"}
+
+	if err := removeMembers(memberList, "test"); err != nil {
```
spiffxp, line 25: So now that I've gotten here, stepping back a bit, it would be cool/ideal if this was all structured in such a way that you could write tests that:
- don't interact with the filesystem
- don't call git

The filesystem thing might be more effort than it's worth (mock out walk, pass in a tree).  But the git thing... maybe write a git interface with `add` and `commit` methods, and pass in a mock in these tests?

https://github.com/kubernetes/test-infra/blob/master/prow/repoowners/repoowners_test.go would be an example of unit tests that dealt with files in a tree layout

cmd/remove-members/test/members.txt
```diff
@@ -0,0 +1,2 @@
+johndoe
```
spiffxp, line 1: another test case to consider: a file not named `orgs.yaml` or `teams.yaml` that has a member's name in it

spiffxp (changes requested):
Thank you for doing this! Some comments and suggestions.  

And I apologize in advance if this conflicts with other feedback or comes across as blowing scope from "rewrite this script in go".  I just wrote without looking, with an eye toward some of the other tooling we use here.

Totally happy to drive to whatever consensus others want.

cmd/remove-members/main.go
```diff
@@ -0,0 +1,176 @@
...
+
+		//Mofify the file only if it's not a dry run
+		if dryrun {
+			return true, len(matches), nil
+		}
+
+		updatedContent := re.ReplaceAll(content, []byte(""))
+		if err = ioutil.WriteFile(path, updatedContent, info.Mode()); err != nil {
+			return false, len(matches), err
+		}
```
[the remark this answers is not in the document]
mrbobbytables: We discussed this a bit earlier in the thread^^;; It got hidden under some other replies when loading though.
https://github.com/kubernetes/org/pull/2575#discussion_r595741022

hack/remove-members.go
daemon1024: Adding to whitespaces and comments related changes, The substantial amount of diff is also created due to the difference in order of the yaml properties.
I was trying to parse the yaml files and removing members in a more structured way, I also faced this.

```diff
diff --git a/config/kubernetes-incubator/org.yaml b/config/kubernetes-incubator/org.yaml
index baf4ca76..e2da8a87 100644
--- a/config/kubernetes-incubator/org.yaml
+++ b/config/kubernetes-incubator/org.yaml
@@ -1,10 +1,3 @@
-name: Kubernetes Incubator
-description: Kubernetes Incubator
-default_repository_permission: read
-has_organization_projects: true
-has_repository_projects: true
-members_can_create_repositories: false
-billing_email: github@kubernetes.io
 admins:
 - cblecker
 - fejta
@@ -15,3 +8,10 @@ admins:
 - nikhita
 - spiffxp
 - thelinuxfoundation
+billing_email: github@kubernetes.io
+default_repository_permission: read
+description: Kubernetes Incubator
+has_organization_projects: true
+has_repository_projects: true
+members_can_create_repositories: false
+name: Kubernetes Incubator
```
It seems to be sorting the fields.

I currently implemented this using `kubernetes-sigs/yaml`.

daemon1024:
I updated how the team and org removals are fetched, 
The tool currently
- reads the yaml file
- unmarshals the file
- reads and stores occurrences of the said member in the parsed yaml
- regex finds the occurences
- verifies if regex finds and stored occurences are same
- removes the finds using regex

removing the occurences from the parsed yaml and marshling it back would have been the optimum way but it generates a substantial amount of diff (Ref https://github.com/kubernetes/org/pull/2575#discussion_r595741022)

If this seems ok, I will proceed with updating the tests for the same

cmd/remove-members/main.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	var members = strings.Split(string(file), "\n")
+	return members, nil
+}
+
+//removeMembers walks through the config directory and removes the occurences of the given member name
+func removeMembers(memberList []string, o options) error {
+	for _, member := range memberList {
+		var orgs, teams []string
```
shamsher31, line 72: This will redeclare orgs and teams with each iteration, better to move it outside for loop.

cmd/remove-members/main.go
```diff
@@ -0,0 +1,228 @@
...
+		"orgs":  fetchOrgRemovals(cfg, []string{}, member, path),
+		"teams": fetchTeamRemovals(cfg.Teams, []string{}, member),
+	}
+
+	if len(removals["orgs"])+len(removals["teams"]) > 0 {
+		re := regexp.MustCompile(`(\s+)?- ` + member + `(.*)?`)
+
+		matches := re.FindAllIndex(content, -1)
+
+		//Making Sure count from parsed config and regex matches are same
```
shamsher31, line 147:
```suggestion
		// Making sure count from parsed config and regex matches are same
```

mrbobbytables:
/remove-lifecycle rotten

/cc @palnabarun 
was this going to be integrated into your helper?

PR #2600, opened 2021-03-25
Update SIG Windows Leads

cblecker:
- Adds specific group for leads
- Updates OWNERS_ALIASES

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -7,7 +7,7 @@ aliases:
     - janetkuo
     - kow3ns
     - mattfarina
-    - prydonius
+    - soltysh
   sig-architecture-leads:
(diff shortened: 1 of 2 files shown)


marosset:
Is there a difference between the sig-windows-leads group defined at https://github.com/kubernetes/org/blob/main/OWNERS_ALIASES and the one being added at https://github.com/kubernetes/org/blob/main/config/kubernetes-sigs/sig-windows/teams.yaml?

cblecker:
Yes (unfortunately). The former is allow sig-windows leads to get `/approve` in this repo to manage your groups, then the latter is the actual github team

marosset:
/lgtm

Ah ok, thanks for the explanation!

PR #2603, opened 2021-03-26
Update kube-state-metrics maintainer list

mrueg:
/assign @lilic 
/assign @mrbobbytables

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1560,6 +1560,7 @@ teams:
     members:
     - brancz
     - lilic
+    - mrueg
     - tariq1890
     privacy: closed
   kubectl-admins:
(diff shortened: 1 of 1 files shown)


config/kubernetes/org.yaml
```diff
@@ -1551,17 +1551,16 @@ teams:
   kube-state-metrics-admins:
     description: Admin access to the kube-state-metrics repo
     members:
-    - andyxning
```
lilic, line 1554: FYI @tariq1890 @brancz as andyxning was added to emeritus-approvers, so removing admin and write access makes sense. We can always add andyxning later on again if they decide to contribute more!

lilic (commented):
lgtm from my side 🎉 

Thanks and welcome Manuel!

config/kubernetes/org.yaml
tariq1890: Sounds good to me!

PR #2604, opened 2021-03-26
Update Kubernetes Code of Conduct Committee names

karenhchu:
updating CoCC member list with new member names


tpepper:
The additions need to be alphabetical it appears, per the test failure.

PR #2620, opened 2021-04-06
kubernetes-csi: pohly as admin of csi-driver-host-path

pohly:
This will make it simpler to release new versions of that component.


nikhita:
/approve

Leaving lgtm to @msau42

PR #2626, opened 2021-04-09
sig-storage: create team for use in image publishing job config

pohly:
We occassionally need to rerun an image build job for a tagged release
because QEMU sometimes is flaky. Members of this team will be able to
trigger that themselves instead of having to track down someone from
test-infra.


pohly:
The corresponding job change is in https://github.com/kubernetes/test-infra/pull/21734

It should be merged next.

Issue #2317, opened 2020-11-06
REQUEST: New membership for tam7t


tam7t:
### GitHub Username
@tam7t

### Organization you are requesting membership in
@kubernetes-sigs

### Requirements
- [x] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [x] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [x] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [x] I am actively contributing to 1 or more Kubernetes subprojects
- [x] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [x] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
@aramase 
@ritazh 
@jimangel

### List of contributions to the Kubernetes project
- PRs reviewed / authored
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/266
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/267
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/280
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/340
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/371
- Issues responded to
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/243
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/256
- SIG projects I am involved with
  - https://github.com/kubernetes-sigs/secrets-store-csi-driver

aramase:
+1 for @tam7t's membership

Tommy has been actively contributing to the Secrets Store CSI Driver project

ritazh:
+1 🎉 

Thank you for your contribution on the sig-auth subproject: https://github.com/kubernetes-sigs/secrets-store-csi-driver

savitharaghunathan:
Hi, @tam7t 👋. Thanks for all your contributions 🎉 

According to the [sponsor requirements](https://github.com/kubernetes/community/blob/master/github-management/new-membership-procedure.md#sponsor-requirements), it requires the sponsors to be from different companies. Can you add another sponsor from another company to maintain integration across the community? @ritazh & @aramase's GH profile has listed Microsoft as their employer. Thanks!

tam7t:
[...]

PR #2637, opened 2021-04-14
add tam7t to kubernetes org

tam7t:
I am already a member of kubernetes-sig (https://github.com/kubernetes/org/issues/2317) and need kubernetes membership for prow and image promotion.

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1010,6 +1010,7 @@ members:
 - szuecs
 - tabbysable
 - tallclair
+- tam7t
 - tamalsaha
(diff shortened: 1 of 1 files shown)


mrbobbytables:
The current org membership policy doesn't translate from kubernetes-sigs to kubernetes, the inverse is true...if you're a member of kubernetes you can become a member of the other orgs, but not the other way around.

This is policy is getting revisited now that there has been a significant of org cleanup, but for now you should still open a GH issue with the request. >_<

mrbobbytables:
fixes: #2638 
/lgtm
/approve
/hold cancel

Issue #2678, opened 2021-04-29
New Repo kubernetes-sigs/network-policy-api


astoycos:
### New Repo, Staging Repo, or migrate existing

new repository

### Requested name for new repository
network-policy-api

### Which Organization should it reside
kubernetes-sigs

### If not a staging repo, who should have admin access
astoycos, abhiraut, rikatz

### If not a staging repo, who should have write access
astoycos, abhiraut, rikatz

### If not a staging repo, who should be listed as approvers in OWNERS
astoycos, abhiraut, rikatz

### If not a staging repo, who should be listed in SECURITY_CONTACTS
astoycos, abhiraut, rikatz

### What should the repo description be

This sig-network sub-project addresses further work involving Kubernetes network security beyond the initial NetworkPolicy resource

### What SIG and subproject does this fall under in sigs.yaml

sig-network

### Approvals

thockin, caseydavenport, dcbw, aojea

### Additional context for request

closes [#5747](https://github.com/kubernetes/community/issues/5747)

Any additional information or context to describe the request:

After much discussion in the [Network Policy API Meeting](https://docs.google.com/document/d/1AtWQy2fNa4qXRag9cCp5_HsefD7bxKe3ea2RPn8jnSs), on the [KEP for Cluster Scoped Network Policy](https://github.com/kubernetes/enhancements/pull/2522/files#r600361541)  and in slack([dis 1](https://kubernetes.slack.com/archives/C01CWSHQWQJ/p1619120255062800) [dis 2](https://kubernetes.slack.com/archives/C09QYUH5W/p1618570463278100)) we (the sig-network-policy subgroup within [Sig-Network](https://github.com/kubernetes/community/tree/master/sig-network)) have decided that the best way to iterate on Network Policy, such as with the cluster-scoped network policy and a possible Network Policy V2, is to create a repo which will serve to house CRDs supported by the group.  This follows the same pattern completed  by the [Kubernetes Gateway API](https://github.com/kubernetes-sigs/gateway-api).  It has also been agreed upon that this repo should be called `network-policy-api` as it relates to future work involving k8's network security.   

Putting the project outside of `kubernetes/kubernetes` will allow for faster iterations and greater design freedom for new objects related to kubernetes networking security 

This issue serves to open up further discussion from the greater Sig-Network group around how the internals (CRD groups, api approval process etc should work) and to hopefully get permission to create the new repository.

caseydavenport:
/lgtm

thanks!

[...]

PR #2697, opened 2021-05-10
Add GitHub teams for network-policy-api

nikhita:
Ref: https://github.com/kubernetes/org/issues/2678

Also adds @abhiraut to @kubernetes-sigs, they are already a member of @kubernetes. 

/assign @palnabarun

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -18,6 +18,7 @@ admins:
 members:
 - aaroniscode
 - aaronlevy
+- abhiraut
 - adelina-t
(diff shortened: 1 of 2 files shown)


config/kubernetes-sigs/org.yaml
```diff
@@ -18,6 +18,7 @@ admins:
 members:
 - aaroniscode
 - aaronlevy
+- abhiraut
```
abhiraut, line 21: @nikhita seems like I received the invitation to join to repo while I was on PTO and the invitation has now expired. How do I receive another invitation? Thanks

PR #2726, opened 2021-05-20
Add prameshj and swetharepakula into ingress-gce maintainer

freehan:
/assign bowei@

/cc prameshj swetharepakula

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1521,7 +1521,9 @@ teams:
     - bowei
     - freehan
     - MrHohn
+    - prameshj
     - rramkumar1
+    - swetharepakula
     - thockin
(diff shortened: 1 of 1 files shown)


config/kubernetes/org.yaml
```diff
@@ -1523,6 +1523,8 @@ teams:
     - MrHohn
     - rramkumar1
     - thockin
+    - prameshj
+    - swetharepakula
```
palnabarun, line 1527: Can you please ensure that the members list is sorted after the addition?

palnabarun (changes requested):

config/kubernetes/org.yaml
freehan: ack

PR #2753, opened 2021-06-01
Add myself as ingress-nginx admin

rikatz:
Adding myself as ingress-nginx admin to create the branch necessary for v1 migration.


palnabarun:
/cc @aledbf @bowei @justinsb 
(for an lgtm)

PR #2800, opened 2021-06-22
sig-instrumentation: Add lilic to approvers

brancz:
@logicalhan

### DIFF
--- a/config/kubernetes/sig-instrumentation/teams.yaml
+++ b/config/kubernetes/sig-instrumentation/teams.yaml
@@ -16,6 +16,7 @@ teams:
     - logicalhan
     - RainbowMango
     - serathius
+    - lilic
     privacy: closed
   sig-instrumentation-members:
     description: ""


logicalhan (approved):
/lgtm
/approve

lilic:
Thanks! 🙇‍♀️  :)

ehashman:
List needs to be alphabetized, which is why CI is failing.

@lilic is not an approver in k/k (where this list was pulled from---these are approvers in the kubernetes/ org) but is in kubernetes-sigs as KSM subproject owner. Should we be adding her here or to a kubernetes-sigs team?

I guess technically KSM is still in the kubernetes org but it's not supposed to be... :)

/hold
for discussion

brancz:
This has nothing to do with kube-state-metrics, @lilic has reviewed plenty of instrumentation-related PRs and I want her to be able to approve any sig instrumentation-related PRs, because I trust whatever review she does, and it distributes the load. I don't know if this PR does that, but it's what I found searching through the org, if this PR does not reflect my intention, can you point out what needs to be different?

ehashman:
https://github.com/kubernetes/kubernetes/blob/6dd9deea3daa98e0003878544372b2c1b9e0e03d/OWNERS_ALIASES#L319-L336

Lili currently isn't listed as either an instrumentation reviewer or approver. We should fix that first before adding her to the GitHub team.

ehashman:
So, to be clear, this PR doesn't actually add any rights or responsibilities, it would just add another user to the GitHub team. I think as proposed it's a bit confusing for the reasons stated above. If we want to make OWNERS files changes, we should be following the [community membership guidelines](https://github.com/kubernetes/community/blob/master/community-membership.md).

I would also like to see @lilic become an OWNER in k/k if that is something she's interested in doing. If that's the case, I think the next steps here are:

- add @lilic as a reviewer to k/k right now
- follow up with [approver access in 3mo](https://github.com/kubernetes/community/blob/master/community-membership.md#approver)

I have great respect and appreciation for @lilic's work, and I would be delighted for you to take on more responsibility in SIG Instrumentation.

brancz:
For the record, Lili has been active in sig instrumentation for well over two years. I don't think us missing out on adding her to be a reviewer long ago is totally our mistake, and I think it's wrong not to acknowledge that mistake.

Issue #2812, opened 2021-07-01
Automate org membership updates


nikhita:
Today, GitHub admins and [new membership coordinators](https://github.com/kubernetes/community/tree/master/github-management#new-membership-coordinator):
1. periodically look through membership requests
2. verify if the sponsors satisfy the required criteria (not from the same company)
3. manually add the GitHub usernames to the config files one-by-one

Both GitHub Admins and the NMCs have constrained bandwidth so we should look into automating some of these.

While 1 and 2 can't really be automated because there is no way to easily verify a sponsor's employer, we should look into automating 3.

Some ideas from an old discussion with @palnabarun and @mrbobbytables:

```
add <gh handle> -org kubernetes -org kubernetes-sigs

add <gh handle> -team kubernetes/maintainers
```

Prior art for removing members in:
1. https://github.com/kubernetes/org/blob/main/hack/remove-members.sh (bash)
2. https://github.com/kubernetes/org/pull/2575 (go)

nikhita:
/assign @palnabarun 
because you were interested in working on this. if you don't have the bandwidth, please unassign and we'll mark this as help-wanted

RinkiyaKeDad:
For 2, I know we can't do it for every case but if a sponsor has specified their company on their github we could do something like 
```
curl https://api.github.com/users/RinkiyaKeDad | jq .company
```
and check that the companies of the sponsors aren't the same.

palnabarun:
@RinkiyaKeDad -- that's a great idea! 👍🏽 

However, the caveat is that sometimes capitalization or the way someone has mentioned the company in that field can make the process cumbersome. We can try stripping special characters and asserting to lowercase, but would still require a level of human supervision.

RinkiyaKeDad:
What I was initially thinking was that we could have the lowercase converted company names be commented on the issue request for the membership but I now realize this wouldn't be of much help since it would save the reviewer at max two clicks 🙂

PR #2822, opened 2021-07-13
Implement CLI to add members to org configurations

palnabarun:
Fixes: #2812

### What is being introduced?

- `korg` cli to act as a toolkit for operating on organization configuration
- `korg add` subcommand to add users to orgs/teams
- A small make recipe to install the binary

This is a proof of concept implementation with everything put into a single package for simplicity. In future, the functionality can be split into cmd/ and pkg/. Some utilities will need unit tests to ensure behaviour remains consistent with time.

#### One-time rejig of the org manifests

Our YAML marshaller/unmarshaller doesn't preserve a custom field order in YAML files that it reads. There is a one-time diff to reformat the config files. Once that is done, the order of fields is consistent across runs. A diff can be found [here][one-time-diff].

#### Comments are not preserved

The sigs.k8s.io/yaml library doesn't parse comments. As a result, when these files are modified and save, the comments go away. For example, as seen in this [diff][comment-diff]. There are around [N instances][comment-instances] of such comments.

#### Incomplete support for nested teams

GitHub supports teams inside teams. Currently, the implementation of `korg add` doesn't support it yet. Once this POC goes through one pass of reviews, I will add support for that as well.

[one-time-diff]: https://gist.github.com/palnabarun/3af6299f9f1f97064800baacda7cd07f#file-one-time-diff
[comment-diff]: https://gist.github.com/palnabarun/3af6299f9f1f97064800baacda7cd07f#file-comment-diff
[comment-instances]: https://gist.github.com/palnabarun/3af6299f9f1f97064800baacda7cd07f#file-comment-instances-diff


cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	commit, err := w.Commit(message, &git.CommitOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to commit changes: %s", err)
+	}
+
+	_, err = r.CommitObject(commit)
+	if err != nil {
+		return fmt.Errorf("unable to write commit object to repo: %s", err)
```
nikhita, line 117: `r.CommitObject` just returns the hash for the commit, right? IIUC it doesn't actually write the commit. The step above (`w.Commit`) does it.

cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+	for _, e := range slice {
+		if key == e {
+			return true
+		}
+	}
+
+	return false
+}
+
+func validateOrgs(orgs []string) (valid []string, invalid []string, invalidPresent bool) {
```
nikhita, line 41: Is there a reason why we return the `valid` orgs? It looks like they aren't being used anywhere. :thinking:

cmd/korg/korg.go
```diff
@@ -0,0 +1,248 @@
...
+	for _, org := range options.Orgs {
+		fmt.Printf("adding %s to %s org\n", username, org)
+		relativeConfigPath := fmt.Sprintf(orgConfigPathFormat, org)
+		configPath := filepath.Join(options.RepoRoot, relativeConfigPath)
+		config, err := readConfig(configPath)
+		if err != nil {
+			return fmt.Errorf("reading config: %s", err)
+		}
+
+		if stringInSlice(config.Members, username) {
```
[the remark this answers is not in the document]
nikhita: Additionally, we'll need to ignore case-sensitivity too. For example, right now `Nikhita` and `nikhita` are treated as different by `stringInSlice` but GitHub treats both as the same username.

cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+	lj := strings.ToLower(s[j])
+
+	return sort.StringsAreSorted([]string{li, lj})
+}
+
+func caseAgnosticSort(arr []string) {
+	sort.Sort(caseAgnostic(arr))
+}
+
+func parseTeam(team string) (string, string, string, error) {
```
nikhita, line 147: Can you add a comment here to document what this function returns?

cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+	parts := strings.Split(team, "/")
+	if len(parts) < 2 || len(parts) > 3 {
+		return "", "", "", fmt.Errorf("invalid team: %s", team)
+	}
+
+	org := parts[0]
+	if !stringInSlice(validOrgs, org) {
+		return "", "", "", fmt.Errorf("invalid team: %s", team)
```
nikhita, line 155: typo...invalid org?

cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+		team = parts[2]
+	}
+
+	return org, group, team, nil
+
+}
+
+func userInOrg(username string, org string, options Options) bool {
+	configPath := filepath.Join(options.RepoRoot, fmt.Sprintf(orgConfigPathFormat, org))
+	config, err := readConfig(configPath)
```
nikhita, line 173: Looks like we need to do the "is user in the org" check for both adding a user to an org and a team. Can we instead read the config once before we call `AddMemberToOrgs` and `AddMemberToTeams` instead of calling it twice?

nikhita (commented):
Left some suggestions but love it overall! :heart_eyes: 

A couple of more questions about formatting:

- It looks like a diff is generated while yaml (un)marshaling due to sorting keys, whitespace changes. Can you also update the config files with these changes so the next time someone runs this tool the only diff is the username update?
- Currently, comments are lost when the yaml file is marshalled. Can we use https://github.com/go-yaml/yaml/issues/132 for preserving comments?

mrbobbytables (commented):
@nikhita covered things really well
The big thing I do consider a blocker is the preservation of the comments.  Theres a significant amount of metadata there that people use. =/

palnabarun:
I have gone through approaches that we can take to parse YAML comments and write them back. In order to do that we would need to change all structures describing org and team config to store comment data and there is no workaround to that. I did some experimentation here: https://gist.github.com/palnabarun/5e773f8e9e62b467e7cf12affe6591f1.

[This answer][so] on Stackoverflow beautifully describes the process of loading and dumping YAML files to/from native data structures. The answer also points out that using YAML comments for storing meaningful is not the right approach since YAML is supposed to be a transient data format.

Since parsing YAML comments is deemed to be a blocker, I had a look at the number of instances where YAML comments are used to annotate the config. The results are at https://gist.github.com/palnabarun/1fe8b4570607586c391f9190132212be.
Only SIG Docs, SIG Release, SIG Scalability and SIG Architecture Enhancements subproject use YAML comments. SIG Storage uses comments for describing the role of a particular team which I feel could be written in their k/community README or some other documentation.

Because of the above, I feel that spending time on modifying existing data structures to parse YAML comments will provide less value compared to the time invested in the feature.

What I propose is:
- Decouple adding members to org and teams as separate concerns so that the work of NMCs is unblocked
- Work on adding teams support to `korg add` later

@nikhita @mrbobbytables WDYT?

[so]: https://stackoverflow.com/a/60891175

cc: @MadhavJivrajani

palnabarun:
I will refactor this PR to split out adding members to org and teams.

cmd/korg/korg.go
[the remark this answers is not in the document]
palnabarun: Both are taken care of now.

cmd/korg/utils.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (s caseAgnostic) Less(i, j int) bool {
+	li := strings.ToLower(s[i])
+	lj := strings.ToLower(s[j])
+
+	return sort.StringsAreSorted([]string{li, lj})
+}
+
+func caseAgnosticSort(arr []string) {
+	sort.Sort(caseAgnostic(arr))
```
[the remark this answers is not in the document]
palnabarun: Done!

cmd/korg/utils.go
palnabarun: 👍🏽

cmd/korg/utils.go
palnabarun: Simplified the implementation now. Valid orgs are not returned anymore.

cmd/korg/utils.go
palnabarun: Removed the team related implementation for now.

cmd/korg/utils.go
palnabarun: Removed the team related implementation for now.

cmd/korg/utils.go
palnabarun: I agree to the suggestion. Reading the config at the start of the command and reusing it makes more sense.

Since the team implementation is deferred to later, we don't need to cache the config in memory for now.

palnabarun:
Based on our discussions, I have reduced the scope of this PR to include only additions to the org and deferred the addition of users to teams to a future timeline.

I have also included the changes to the configuration files in this PR.

nikhita (approved):
Just tried this command and it's AMAZING!! :heart_eyes: 

Thanks for working on this!

/lgtm
/approve

/hold
I'm good with merging this as-is. Adding a hold in case you want more reviews.

nikhita:
let's merge this?

PR #2830, opened 2021-07-15
Add jihoon-seo to sig-docs-ko-reviews

jihoon-seo:
Add @jihoon-seo to sig-docs-ko-reviews.

related with https://github.com/kubernetes/website/pull/28948


seokho-son:
+1 @jihoon-seo is very active contributor for Korean localization of k/website. I am happy to work with @jihoon-seo

PR #2852, opened 2021-07-30
add cluster/ approvers to kubernetes-maintainers

spiffxp:
we are looking into using blockade for PRs that touch `cluster/` in a
similar manner that is used for kubernetes/release - someone will need
to manually remove the blockade label before a PR is allowed to merge

these two cluster/ approvers aren't part of the kubernetes-maintainers
team and so would not have the power to do so


dims:
LGTM

PR #2868, opened 2021-08-05
Add chrishenzie as maintainer of kubernetes-csi repos

chrishenzie:
So that I can perform releases for these CSI sidecars.

/assign @msau42


msau42:
/lgtm

Chris is helping out with the releases for numerous csi components

Issue #2222, opened 2020-09-24
Changing default branch names to `main`


justaugustus:
### Organization or repository

All

### Users affected

Global

### Describe the issue

This is a long-term tracking issue for renaming the default branch on repositories from `master` to `main`. Consider this community tracking for the discussion in https://github.com/github/renaming.

As of today (9/24):
- Starting on October 1, 2020: newly-created repositories will default to `main`: https://github.com/github/renaming#on-october-1-2020-newly-created-repositories-will-default-to-main
- If you are a maintainer of an existing repo, do **NOT** rename your existing default branch
  - We're waiting for later in the year when the change will be less disruptive for consumers: https://github.com/github/renaming#later-this-year-seamless-move-for-existing-repositories-

cc: @kubernetes/owners @kubernetes/wg-naming 
/area github-admin
/sig contributor-experience
/wg namin

celestehorgan:
@justaugustus – Do you think this merits a communication to k/dev?

justaugustus:
/assign
cc: @liggitt

justaugustus:
> @justaugustus – Do you think this merits a communication to k/dev?

@nikhita's initial note to k-dev should suffice at this stage: https://groups.google.com/g/kubernetes-dev/c/jvRPIscaDek/m/1KWEUYXDBQAJ.

liggitt:
The only impact I can think of for *new* repos would be that the git.k8s.io redirector (`http://git.k8s.io/$repo/$path` going to `https://github.com/kubernetes/$repo/tree/master/$path`, e.g. http://git.k8s.io/community/sig-apps) would not work properly.

nikhita:
> Starting on October 1, 2020: newly-created repositories will default to main: https://github.com/github/renaming#on-october-1-2020-newly-created-repositories-will-default-to-main

Just to ensure we are on the same page - when we create a new repo in a Kubernetes org after Oct 1, we will set the default branch to `master` for now, correct? IIUC we'll move _all_ our repos to the `main` default branch once GitHub solves the issues listed here - https://github.com/github/renaming#later-this-year-seamless-move-for-existing-repositories-

nikhita:
xref https://github.com/kubernetes/website/issues/21749 - k/website issue about changing their default branch to `live`

mrbobbytables:
[...]

PR #2886, opened 2021-08-09
hack: add default-branches-report.sh

spiffxp:
Related:
- part of: https://github.com/kubernetes/org/issues/2222

outputs a github-flavored markdown checklist of repos per org, with
checkboxes checked if their default branch is named something other than
'master'

run with no args to target the orgs managed by the kubernetes
github-management subproject


nikhita (approved):
/lgtm
/approve

PR #2893, opened 2021-08-16
Add sig-auth lead

ritazh:
Signed-off-by: Rita Zhang <rita.z.zhang@gmail.com>

Context:
https://github.com/kubernetes/community/pull/5728

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -17,6 +17,7 @@ aliases:
     - enj
     - liggitt
     - mikedanese
+    - ritazh
     - tallclair
   sig-autoscaling-leads:
     - gjtempleton


OWNERS_ALIASES
```diff
@@ -17,6 +17,7 @@ aliases:
     - enj
     - liggitt
     - mikedanese
+    - ritazh
     - tallclair
```
fejta, line 21: Remove?

ritazh: @fejta Looks like https://github.com/kubernetes/org/pull/2894/files#diff-4985b733677adf9dda6b5187397d4700868248ef646d64aecfb66c1ced575499L20 has addressed this

PR #3007, opened 2021-09-24
use make

spiffxp:
I grew tired of recompiling protobuf to see if our YAML-based config was sorted

Adds the following human-facing targets:

- `merge`: builds and installs the merge binary
- `config`: creates a merged config using the merge binary
- `test`: tests the merged config using go test
- `verify`: runs hack/verify.sh
- `peribolos`: clones a local copy of kubernetes/test-infra and installs
  the peribolos binary
- `deploy`: runs periobolos to deploy the merged config if tests pass
- `clean`: remove all intermediary configs and binaries

Overrideable env vars:

- `GITHUB_TOKEN_PATH`: value for `--github-token-path` passed to `peribolos`
- `TEST_INFRA_PATH`: set to location of a pre-existing clone of
  kubernetes/test-infra to avoid cloning a temporary copy

One way this differs from bazel is that we're not necessarily getting
the latest version of test-infra each time we run `make deploy` but I
don't foresee peribolos changing that frequently, and a `make clean` is
enough to get us back to "latest" state.


hack/boilerplate/verify_boilerplate.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
spiffxp, line 1: A note here: I didn't make any of these edits myself.  I merely downloaded the latest copy of the script from kubernetes/repo-infra

mrbobbytables:
/lgtm
/hold
for @cblecker to weigh in 👍

cblecker (approved):

PR #3008, opened 2021-09-24
config/kubernetes: sig-k8s-infra team updates

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/org/pull/3007
- Followup to: https://github.com/kubernetes/test-infra/pull/23745

This is mostly to exercise jobs to confirm they all still work

PR #3009, opened 2021-09-24
remove bazel-related files

spiffxp:
Related:
- Followup to: https://github.com/kubernetes/org/pull/3007
- Followup to: https://github.com/kubernetes/test-infra/pull/23745
- Waiting on: https://github.com/kubernetes/org/pull/3008


spiffxp:
/hold
Not yet

spiffxp:
The reason I held is that `make` is not well suited to plumbing through arbitrary flags.  So the postsubmits were dropping the duplicate `--github-endpoint` flags, resulting in peribolos not using ghproxy. 

I gave up trying to use `make` as the entrypoint for all CI jobs:
- https://github.com/kubernetes/org/pull/3013
- https://github.com/kubernetes/test-infra/pull/23766

Currently watching the pot boil on https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-org-peribolos/1442502192983445504 to confirm ghproxy is being used again

PR #3010, opened 2021-09-24
make: fix make deploy

spiffxp:
Trying to avoid changing the job config args, which are passed in as `--foo=bar`. This doesn't work with the approach taken in kubernetes/k8s.io to pass flags through to a make target.  So use a different approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
 # available for override
 GITHUB_TOKEN_PATH ?=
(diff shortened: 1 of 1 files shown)


spiffxp:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-org-peribolos/1441540041988902912 is still running, which is way way longer than usual

```
/home/prow/go/src/github.com/kubernetes/org/_output/bin/peribolos \
	--config-path /home/prow/go/src/github.com/kubernetes/org/_output/gen-config.yaml \
	--fix-org \
	--fix-org-members \
	--fix-teams \
	--fix-team-members \
	 \
	 --required-admins=cblecker  --required-admins=fejta  --required-admins=idvoretskyi  --required-admins=mrbobbytables  --required-admins=nikhita  --required-admins=spiffxp \
	--tokens=1200 --github-token-path=/etc/github-token/oauth --github-endpoint=https://api.github.com --confirm
```

Turns out the hack to pass flags doesn't quite work: two `--github-endpoint` flags enter, one flag leaves.  And unfortunately it's not the one pointing to the ghproxy endpoint.  So the job is getting super throttled on token usage.

Much as I wanted to do this cleanly without changing job configs, I'm going to need to: add a proper script that can be called with flags, and then update jobs to use that script.

We're not broken, but we're slowed to a crawl.  I will put up the PRs to fix forward but am totally open to rolling back if that's what we'd prefer to do

spiffxp:
OK it turns out passing `--foo bar` `--foo baz` will allow both flags though, so I'm just going to change the job config and call it for the week

PR #3024, opened 2021-10-01
SIG-CLI: KnVerey to lead, pwittrock to emeritus

KnVerey:
Part of kubernetes/community#6102

I'm not sure about some of the changes in this one. I added myself to all the SIG-related Github teams, on the assumption those are for getting pinged on relevant issues. But for subproject teams, I only added myself to cli-experimental admins (because that should probably come with Kustomize admin--our docs are there). Should I be on any other subproject team? @pwittrock do you want to be removed from any of these places?

/cc @eddiezane @pwittrock @soltysh @seans3

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -22,7 +22,7 @@ aliases:
     - mwielgus
   sig-cli-leads:
     - eddiezane
-    - pwittrock
+    - KnVerey
     - seans3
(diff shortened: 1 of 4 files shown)


pwittrock:
@ spiffxp mind blessing Katrina's ascension?

cblecker (approved):

PR #3097, opened 2021-11-15
add jonyhy96 as k8s-sig member

jonyhy96:
I already is k8s member and want to apply for sig member.


savitharaghunathan (commented):
/lgtm
/approve

PR #3129, opened 2021-12-06
add andyzhangx to csi-driver-iscsi admin

andyzhangx:
trying to make an alpha release first.

### DIFF
--- a/config/kubernetes-csi/org.yaml
+++ b/config/kubernetes-csi/org.yaml
@@ -149,6 +149,7 @@ teams:
   csi-driver-iscsi-admins:
     description: Admin access to csi-driver-iscsi repo
     members:
+    - andyzhangx
     - jsafrane
     - msau42
     - saad-ali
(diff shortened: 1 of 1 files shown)


palnabarun:
/cc @xing-yang @msau42 @saad-ali 
(for an lgtm)

config/kubernetes-csi/org.yaml
```diff
@@ -233,6 +235,7 @@ teams:
   csi-lib-iscsi-admins:
     description: Admin access to csi-lib-iscsi repo
     members:
+    - andyzhangx
```
msau42, line 238: I believe we no longer maintain the library repo anymore. Instead we pulled it into the csi-driver-iscsi repo.

andyzhangx: that's ok, mainly I would like to maintain csi-driver-iscsi repo

palnabarun:
Approving since an existing admin of the repo (@msau42) had LGTM'ed the PR.

/approve

PR #3227, opened 2022-02-01
Onboard/Offboard Mo/Swamy from SRC

tallclair:
Follow up to kubernetes/committee-security-response#141

/assign @cjcullen @enj @swamymsft


mrbobbytables:
/approve

the OWNERS_ALIASES needs to be updated, but I think theres a general sync that needs to happen from k/community where we can do that one 👍

PR #3241, opened 2022-02-09
Use new dev email in membership request

chases2:
The current group is being sunset in favor of the new group. This changes the link to point directly to it.


mrbobbytables (commented):
Thanks! I thought we changed this when we did our big search/replace earlier this year >_<

/lgtm
/approve

PR #3253, opened 2022-02-14
add jyotimahapatra to kubernetes-sig

jyotimahapatra:
I'm already a part of k8s org https://github.com/kubernetes/org/blob/main/config/kubernetes/org.yaml#L622

Creating the PR based on the recommendation https://github.com/kubernetes/org/issues/3249#issuecomment-1038679567

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -344,6 +344,7 @@ members:
 - juanvallejo
 - justaugustus
 - justinsb
+- jyotimahapatra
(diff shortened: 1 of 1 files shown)


jyotimahapatra:
@palnabarun Are there more steps before the membership reflects . I see an error for `/verify-owners` in here https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/429#issuecomment-1039709655

palnabarun:
You should get a membership invite notification which you need to accept and then only you the check would pass on that PR.

PR #3271, opened 2022-02-16
Add chaodaiG as k8s-sig member

chaodaiG:
fixes: https://github.com/kubernetes/org/issues/3270

chaodaiG is already a member of kubernetes

PR #3272, opened 2022-02-17
Adding michelle192837 as a k8s-sig member.

michelle192837:
As suggested in https://github.com/kubernetes/org/pull/3271#issuecomment-1042061604 ^^


savitharaghunathan (approved):
/lgtm
/approve

PR #3276, opened 2022-02-17
Add mpherman2 as k8s-sig member

mpherman2:
As suggested in https://github.com/kubernetes/org/pull/3271#issuecomment-1042061604


savitharaghunathan (commented):
/lgtm
/approve

PR #3391, opened 2022-04-26
add andyzhangx as maintainer & admin of local-static-provisioner

andyzhangx:
/assign @cofyc


cofyc:
cc @msau42

palnabarun:
/approve
(since there is  LGTM from another member of the teams being added to)

PR #3408, opened 2022-05-05
Fix alias for new member rdpsin

rdpsin:
Fixes an error where Github alias was wrong. 

Original issue here: https://github.com/kubernetes/org/issues/3372

Signed-off-by: Rishabh Singhvi <rdpsin@amazon.com>


rdpsin:
cc: @nikhita

palnabarun:
@rdpsin -- Thank you for noticing the typo and fixing it! ❤️

PR #3421, opened 2022-05-14
update kubebuilder admins as defined in the project owners

camilamacedo86:
**Description** 
Updating permissions to have the same what is defined in the https://github.com/kubernetes-sigs/kubebuilder/blob/master/OWNERS_ALIASES

### DIFF
--- a/config/kubernetes-sigs/sig-api-machinery/teams.yaml
+++ b/config/kubernetes-sigs/sig-api-machinery/teams.yaml
@@ -104,14 +104,6 @@ teams:
     description: admin access to kubebuilder
     members:
     - camilamacedo86
-    - DirectXMan12
-    - droot
-    - estroz
-    - grodrigues3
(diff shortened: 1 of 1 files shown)


config/kubernetes-sigs/sig-api-machinery/teams.yaml
```diff
@@ -104,14 +104,6 @@ teams:
...
     - camilamacedo86
-    - DirectXMan12
-    - droot
-    - estroz
-    - grodrigues3
-    - Liujingfang1
-    - mengqiy
-    - monopole
-    - pmorie
     - pwittrock
```
camilamacedo86, line 107: @mrbobbytables should we not keep here only who is the project owner?

The following ones I know are folks who may have context on ancient history but are no longer directly involved

  - directxman12
  - droot
  - mengqiy
  - estroz

PR #3433, opened 2022-05-20
Add seanmalloy and ingvagabund to descheduler repo with write access

seanmalloy:
There is currently only one active maintainer with write access to the
descheduler repo. Adding a second maintainer with write access allows
multiple people to help with create git tags and branches for releases.

### DIFF
--- a/config/kubernetes-sigs/sig-scheduling/teams.yaml
+++ b/config/kubernetes-sigs/sig-scheduling/teams.yaml
@@ -24,6 +24,7 @@ teams:
     - damemi
     - k82cn
     - ravisantoshgudimetla
+    - seanmalloy
     privacy: closed
     previously:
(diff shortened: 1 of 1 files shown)


seanmalloy:
@damemi and @ingvagabund let me know if you have any concerns with this change.

/cc @damemi @ingvagabund

damemi (commented):
/approve
would also like to nominate @ingvagabund

seanmalloy:
> would also like to nominate @ingvagabund

I pushed a second commit that adds @ingvagabund with write access.

seanmalloy:
@ahg-g and @Huang-Wei please take a look when you have some time.

/assign @Huang-Wei

Huang-Wei:
/approve

(but not sure ^^ works 😅)

Issue #2770, opened 2021-06-09
REQUEST: New membership for HongjiangHuang


HongjiangHuang:
### GitHub Username
@HongjiangHuang

### Organization you are requesting membership in 

@kubernetes, @kubernetes-client

### Requirements
- [x] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [x] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [x] I have subscribed to the Kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [x] I am actively contributing to 1 or more Kubernetes subprojects
- [x] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [x] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors

@tengqm 
@pacoxu 

### List of contributions to the Kubernetes project
- PRs reviewed / authored
  - https://github.com/kubernetes/website/pull/28041
  - https://github.com/kubernetes/website/pull/28250
  - https://github.com/kubernetes/website/pull/28243
  - https://github.com/kubernetes/website/pull/28242
  - https://github.com/kubernetes/website/pull/28241
  - https://github.com/kubernetes/website/pull/28298
  - https://github.com/kubernetes/website/pull/28304
  - https://github.com/kubernetes/website/pull/28330
  - https://github.com/kubernetes/website/pull/28327
  - https://github.com/kubernetes/website/pull/28328
  - https://github.com/kubernetes/website/pull/28329
  - https://github.com/kubernetes-client/gen/pull/191
  - https://github.com/kubernetes-client/gen/pull/192
  - https://github.com/kubernetes/kubernetes/pull/102245

- Issues responded to
  - https://github.com/kubernetes/website/issues/27673
  - https://github.com/kubernetes/website/issues/28284
  - https://github.com/kubernetes/website/issues/28232
  - https://github.com/kubernetes/website/issues/28230
  - https://github.com/kubernetes/website/issues/28228
  - https://github.com/kubernetes/website/issues/28203

pacoxu:
+1

Thanks for the flake fix try and Chinsese website improvements.
Hope to see more contributions from you!

tengqm:
+1
Looking forward to more contributions from you.

savitharaghunathan:
Hi, @HongjiangHuang 👋  Thank you for all your contributions!

[...]

Issue #2096, opened 2020-08-10
REQUEST: New membership for ZhiFeng1993


ZhiFeng1993:
### GitHub Username
@ZhiFeng1993

### Organization you are requesting membership in
@kubernetes

### Requirements
- [X] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [X] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [X] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [X] I am actively contributing to 1 or more Kubernetes subprojects
- [X] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [X] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- @spiffxp 
- @alejandrox1 

### List of contributions to the Kubernetes project
- PRs reviewed / authored
https://github.com/kubernetes/test-infra/pull/18636
https://github.com/kubernetes/test-infra/pull/18621
https://github.com/kubernetes/test-infra/pull/18620
https://github.com/kubernetes/test-infra/pull/18619
https://github.com/kubernetes/test-infra/pull/18300
https://github.com/kubernetes/test-infra/pull/18103
https://github.com/kubernetes/test-infra/pull/17898
https://github.com/kubernetes/kubernetes/pull/92208
https://github.com/kubernetes/kubernetes/pull/90688
https://github.com/kubernetes/kubernetes/pull/93213
https://github.com/kubernetes/kubernetes/pull/93047
https://github.com/kubernetes/kubernetes/pull/93036
https://github.com/kubernetes/kubernetes/pull/91747
https://github.com/kubernetes/kubernetes/pull/91479
https://github.com/kubernetes/kubernetes/pull/90693
https://github.com/kubernetes/website/pull/22792
https://github.com/kubernetes-sigs/yaml/pull/36
https://github.com/kubernetes/website/pull/21781
https://github.com/kubernetes/website/pull/21467

- Issues responded to
https://github.com/kubernetes/test-infra/issues/18583
https://github.com/kubernetes/test-infra/issues/18579
https://github.com/kubernetes/test-infra/issues/18578
https://github.com/kubernetes/test-infra/issues/18577
https://github.com/kubernetes/kubernetes/issues/81245
https://github.com/kubernetes/kubernetes/issues/92911
https://github.com/kubernetes/kubernetes/issues/91042
[...]

Issue #2058, opened 2020-07-22
REQUEST: New membership for sattarfeizollahibarough


sattarfeizollahibarough:
### GitHub Username
sattarfeizollahibarough

### Organization you are requesting membership in
@kubernetes

### Requirements
- [x] I have reviewed the community membership guidelines (https://git.k8s.io/community/community-membership.md)
- [x] I have enabled 2FA on my GitHub account (https://github.com/settings/security)
- [x] I have subscribed to the kubernetes-dev e-mail list (https://groups.google.com/forum/#!forum/kubernetes-dev)
- [x] I am actively contributing to 1 or more Kubernetes subprojects
- [x] I have two sponsors that meet the sponsor requirements listed in the community membership guidelines
- [x] I have spoken to my sponsors ahead of this application, and they have agreed to sponsor my application

### Sponsors
- @zacharysarah 
- @sftim 

### List of contributions to the Kubernetes project
- PRs reviewed / authored
   -https://github.com/kubernetes/org/pull/2051
   -https://github.com/cncf/foundation/pull/101
   -https://github.com/kubernetes/website/pull/22565
   -https://github.com/kubernetes/test-infra/pull/18369
- Issues responded to 
- SIG projects I am involved with
   -sig-docs

sftim:
It's a :+1: from me @sattarfeizollahibarough

sattarfeizollahibarough:
> 
> 
> It's a +1 from me @sattarfeizollahibarough

And I appreciate that :pray: @sftim

zacharysarah:
The org should be `kubernetes` instead of `kubernetes-sigs`, otherwise +1 for our new Persian localization lead.

sftim:
Ah yes, good spot @zacharysarah

sattarfeizollahibarough:
I have modified that. @zacharysarah

sattarfeizollahibarough:
Hi guys, @zacharysarah @sftim 
I hope you are doing well. I am sorry for bugging you. You have approved my membership request kindly, but the request is "open" now. What would be the next step? Thank you in advanced for your response.

nikhita:
@sattarfeizollahibarough that would be on me :) 

Thanks for kickstarting the Persian localization for docs! :100: 

I've created PR #2061 to add you to the Kubernetes Org. Once that gets merged, you should get a membership invite notification.

Welcome to @kubernetes! 🎉

/assign

sattarfeizollahibarough:
[...]

PR #1041, opened 2019-07-24
Add members 2019-07-24

nzoueidi:
Fixes #1032
Fixes #1033
Fixes #1034
Fixes #1035
Fixes #1036
Fixes #1037

/area github-membership

PR #3460, opened 2022-05-31
Update GitHub handles for renamed members

Sea-n:
There are 4 users changed their username, here's the original member request issue:
- #2932
    - @cccswann -> @ccswann-godaddy
- #2770
    - @HongjiangHuang -> @alberthuang24
- #2096
    - @ZhiFeng1993 -> @helenfeng737
- #2058
    - @sattarfeizollahibarough -> @sattarfgh
    - Removed according to [this comment](https://github.com/kubernetes/org/pull/3460#pullrequestreview-990522818).

There are one person cannot be found, but not affected by this PR.
- #1041
	- 404 not found: #1034
	- User profile cannot be open: @kittenking

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1120,7 +1120,6 @@ members:
 - SaranBalaji90
 - saschagrunert
 - SataQiu
-- sattarfeizollahibarough
 - savitharaghunathan
 - sayanchowdhury
 - sbangari


mrbobbytables (changes requested):
Thanks @Sea-n for catching that.

This repo serves as sort of the general log of all memberhsip changes, to make things easily trackable we have a particular format we try and stick to for adding/removing org memberships with 1 change per commit and the message formatted like
```
Add <person> to <org>
Remove <person> from <org>
Rename <old-person> -> <new person>
```
I would also remove sattarfgh, they haven't been active in some time and would be slated to be removed in the next org audit.

If you want to update your PR with those changes, happy to approve. Otherwise I'm happy to open a follow up PR with the parcticular formatting :) It's something our GitHub admins/new member coordinators are familiar with and not commonly known.

mrbobbytables (commented):
awesome, thanks @Sea-n!
/approve

PR #3500, opened 2022-06-13
onboarding the rest of 1.25 release team

cici37:
onboarding the rest of 1.25 release team

PR #3501, opened 2022-06-13
Fix typo, rm krol and add krol3

leonardpahlke:
Signed-off-by: leonardpahlke <leonard.pahlke@googlemail.com>

A typo made in this [PR](https://github.com/kubernetes/org/issues/3478) added @krol and not @krol3

ref: https://github.com/kubernetes/org/pull/3500

/cc @cici37 @reylejano @kcmartin


cici37:
Thanks for addressing it!

PR #3547, opened 2022-07-10
kubernetes-csi: jingxu97 as admin of csi-driver-host-path

xing-yang:
For the releases of csi-driver-host-path


palnabarun:
/lgtm
/approve

(approving from GitHub admin side)

(assuming implicit approval here as @xing-yang is part of the teams being modified)

PR #3574, opened 2022-07-25
update sig-scheduler teams.yaml

Huang-Wei:
Per https://github.com/kubernetes/kubernetes/issues/111392#issuecomment-1194116115, update the outdated teams.yaml in sig-scheduling.

### DIFF
--- a/config/kubernetes/sig-scheduling/teams.yaml
+++ b/config/kubernetes/sig-scheduling/teams.yaml
@@ -2,170 +2,86 @@ teams:
   sig-scheduling-api-reviews:
     description: ""
     members:
-    - aveshagarwal
-    - bsalamat
-    - ConnorDoyle
-    - davidopp
-    - derekwaynecarr
-    - hongchaodeng
-    - ingvagabund
-    - jayunit100
-    - k82cn
-    - kevin-wangzefeng
-    - pigmej
-    - quinton-hoole
-    - resouer
-    - rrati
-    - timothysc
-    - vishh
-    - xiang90
+    - ahg-g
+    - alculquicondor
+    - chendave
+    - damemi
+    - denkensk
+    - Huang-Wei
+    - kerthcet
+    - sanposhiho
     privacy: closed
   sig-scheduling-bugs:
     description: ""
     members:
-    - aveshagarwal
-    - bsalamat
-    - ConnorDoyle
-    - davidopp
-    - derekwaynecarr
-    - gyliu513
-    - hongchaodeng
-    - ingvagabund
-    - jayunit100
-    - k82cn
-    - kevin-wangzefeng
-    - pigmej
-    - quinton-hoole
-    - ravisantoshgudimetla
-    - resouer
-    - rrati
-    - timothysc
-    - vikaschoudhary16
-    - vishh
-    - wanghaoran1988
-    - xiang90
+    - ahg-g
+    - alculquicondor
+    - chendave
+    - damemi
+    - denkensk
+    - Huang-Wei
+    - kerthcet
+    - sanposhiho
     privacy: closed
   sig-scheduling-feature-requests:
     description: ""
     members:
-    - aveshagarwal
-    - bsalamat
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-scheduling/teams.yaml
```diff
@@ -2,170 +2,82 @@ teams:
   sig-scheduling-api-reviews:
```
alculquicondor, line 2: api-reviewers? It's actually just me https://github.com/kubernetes/kubernetes/blob/7b9a20b587de81ed7b885e97f2fd1db1deca548b/OWNERS_ALIASES#L452

config/kubernetes/sig-scheduling/teams.yaml
```diff
@@ -2,170 +2,82 @@ teams:
...
+    - ahg-g
+    - alculquicondor
+    - chendave
+    - damemi
+    - denkensk
+    - Huang-Wei
+    - kerthcet
+    - sanposhiho
     privacy: closed
   sig-scheduling-feature-requests:
```
alculquicondor, line 22: Shouldn't it be just the Chairs and TLs here?

damemi (commented):
do we need all of these different aliases? A lot of them are the same people, so I don't necessarily see the benefit of having `sig-scheduling-bugs`, `sig-scheduling-feature-requests`, `sig-scheduling-pr-reviews`, and `sig-scheduling-misc`. If we're cleaning this up, have we considered consolidating some of these into a generic `@sig-scheduling`?

alculquicondor:
I considered that, but other SIGs have the aliases listed here.

Invoking @mrbobbytables, from sig-contrib-ex for recommendations again. Can we do our own thing? Should other sigs follow our steps?

config/kubernetes/sig-scheduling/teams.yaml
Huang-Wei: This would be just for Github notification purposes, not technically "assigned" to someone. So should be fine IMO.

config/kubernetes/sig-scheduling/teams.yaml
Huang-Wei: Same as above. It's fine to "notify" more ppl about incoming feature requests.

mrbobbytables:
> I considered that, but other SIGs have the aliases listed here.
> 
> Invoking @mrbobbytables, from sig-contrib-ex for recommendations again. Can we do our own thing? Should other sigs follow our steps?

You can do your own thing, these were sort of the "defaults" when setting up a SIG ages ago. If they aren't being used, they aren't serving a purpose. Just be sure to update sigs.yaml with the new teams 👍

config/kubernetes/sig-scheduling/teams.yaml
kerthcet: If appliable, plz add me to the notify list.

denkensk: +1 I am interested in it.

ahg-g:
I would consolidate all of those groups into two only: reviewers and approvers. What do you folks think?

alculquicondor:
And since our approvers are different from reviewers, it might make sense to have `sig-scheduling-all` as well?

kerthcet:
Can we push this forward? @Huang-Wei inspired by https://github.com/kubernetes/website/pull/35042#issuecomment-1216037729

Huang-Wei:
@mrbobbytables could you approve this? As a first step (this PR), we want to cleanup the dated members.

Next, we may simplify the number of aliases as they all point to the same group of people now.

alculquicondor:
@ahg-g also got rejected to apply the label by the bot. So I think we need to move forward with this and the follow up PR.

mrbobbytables (commented):
Sorry for the delay^^;;; between OOTO and a slew of other notifications this slipped through the cracks.

/approve

alculquicondor:
/lgtm
?

PR #3603, opened 2022-08-09
Add dims to test-infra ACL(s)

dims:
Follow up 4c77d274a3e2ddc05221b6c6cff1205ada90f7d6

Lot of folks on test-infra-admins and test-infra-maintainers are gone now! 

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


alvaroaleman:
Why is this needed?

dims:
@alvaroaleman it would be good to have someone who is currently active just in case we land up in trouble and github karma is needed to fix things.

cjwagner (approved):

PR #3626, opened 2022-08-18
Add qbarrand as KMM admin

qbarrand:
Add @qbarrand as a maintainer for [Kernel Module Management](https://github.com/kubernetes-sigs/kernel-module-management) that we are transitioning from https://github.com/qbarrand/oot-operator.


derekwaynecarr:
discussed in sig-node 8/30/2022 call.
/lgtm
/approve

dchen1107:
/lgtm
/approve

Thanks!

PR #3631, opened 2022-08-20
Add JiaoDean mowangdk as alibaba-csi-plugin admin

fredkan:
JiaoDean and mowangdk are working on alibabacloud csi plugin as admin. I'm proposing to add them as proper admin of alibaba-cloud-csi-driver project.


cheftako:
Guessing you have '@' signs in the commit messages and it does not like that. Can you amend the commit to remove those characters?

palnabarun:
@fredkan you would need to amend the first commit of the PR.

Issue #3643, opened 2022-08-29
Netlify site for AdminNetworkPolicy API


astoycos:
### Repository

kubernetes-sigs/network-policy-api

### Project domain

adminnetworkpolicy-api.sigs.k8s.io

### Additional context for request

Following what was [done for the gateway api](https://github.com/kubernetes/org/issues/2506) to ultimately deploy a website for the new adminNetworkPolicy API defined here -> https://github.com/kubernetes-sigs/network-policy-api/pull/33

astoycos:
@mrbobbytables I'm hoping you can help us out here :)... The source for the website is still in a PR https://github.com/kubernetes-sigs/network-policy-api/pull/33 but should be ready to go 

Please let me know if I missed any steps here to get this deployed (I tried to follow what the Gateway API folks did) 

Thanks in Advance!!

astoycos:
In the meantime I also plan to deploy this on GH pages :)

mrbobbytables:
Sorry, between being OOTO and other responsibilities I didn't get to this earlier - netlify has been configured, the site name right now kubernetes-sigs-network-policy-api.netlify.app 

Looking at the netlify.toml file in the root of the repo, it should be updated to be .app instead of .com and site name change to be kubernetes-sigs-network-policy-api - its too long with admin in the name of it for netlify >_<
There are also some build errors from the initial deploy test:
https://app.netlify.com/sites/kubernetes-sigs-network-policy-api/deploys/63233e95521a3a5e969288cc

astoycos:
No problem, I will update the name and investigate the build errors, thanks so much @mrbobbytables! :)

astoycos:
I wonder do I also need to fix this https://github.com/kubernetes/k8s.io/pull/4151/files now or is that just a nice TODO?

mrbobbytables:
That'll need to get updated too for the dns non netlify dns name to work >_<

astoycos:
Site is currently ip at https://kubernetes-sigs-network-policy-api.netlify.app/ And I just pushed an update to the DNS records

astoycos:
Should be GTG now @mrbobbytables Thanks for all the help folks!

cblecker:
I'm doing the Netlify piece right now :)

astoycos:
Awesome, thanks for the help @cblecker!

mrbobbytables:
Thanks @cblecker! I was just about to login myself lol

cblecker:
[...]

PR #3644, opened 2022-08-29
Add astoycos to NPAPI admins

astoycos:
astoycos is an owner in the
adminNetworkPolicy API repo
and should be added here as
well.

Related to https://github.com/kubernetes/org/issues/3643


astoycos:
cc @rikatz @abhiraut

rikatz:
Lgtm
Andrew is the fact npapi admin and I guess I can even stepdown on this :)

rikatz (approved):

PR #3650, opened 2022-08-31
Rename kernel-module-management teams

cblecker:
Renames the kernel-module-management teams to their correct name.


cblecker:
/hold
need to make sure #3649 gets rolled out first

cblecker:
/hold cancel
verified last roll out complete: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-org-peribolos/1565012281685708800

PR #3653, opened 2022-09-01
update sig-arch enhancements members

kikisdeliveryservice:
Removing old RT Enhancements members and add 1.26 rt enhancements lead.

### DIFF
--- a/config/kubernetes/sig-architecture/teams.yaml
+++ b/config/kubernetes/sig-architecture/teams.yaml
@@ -38,17 +38,12 @@ teams:
     maintainers:
     - mrbobbytables # subproject owner
     members:
-    - gracenng # 1.22 Enhancements Shadow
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice:
Pending PR creating group: https://github.com/kubernetes/k8s.io/pull/4176

leonardpahlke:
@kikisdeliveryservice the referenced PR merged, we should be good to go with this one, right

palnabarun:
I don't think this PR should be blocked on other PRs. The RT Enhancements members in the existing team are old members and can be removed.

johnbelamaric:
/approve

sheesh

kikisdeliveryservice:
the alias doesn't work here with the repo setup. added rt enhancemnets lead manually.

Issue #3605, opened 2022-08-11
REQUEST: Migrate oci-proxy


BobyMCbobs:
### New repo, staging repo, or migrate existing

kubernetes-sigs/oci-proxy

### Is it a staging repo?

no

### Requested name for new repository

registry.k8s.io

### Which Organization should it reside

kubernetes

### Who should have admin access?

ameukam, bentheelder, dims, thockin

### Who should have write access?

ameukam, bentheelder, dims, thockin

### Who should be listed as approvers in OWNERS?

ameukam, bentheelder, dims, thockin

### Who should be listed in SECURITY_CONTACTS?

ameukam, bentheelder, dims, thockin

### What should the repo description be?

A custom HTTP redirector serving Kubernetes container image blobs from existing vendor supported container registries.

### What SIG and subproject does this fall under?

sig-k8s-infra

### Please provide references to appropriate approval for this new repo

https://github.com/kubernetes-sigs/oci-proxy/issues/94#issuecomment-1211308265

dims:
+1

ameukam:
+1

Riaankl:
/assign @nikhita @mrbobbytables 

Is there any "next steps" that we should disucss or start working on?

Riaankl:
This issue was discussed today in SIG K8s Infra meeting.

According to the [guidance given](https://github.com/kubernetes-sigs/oci-proxy/issues/94#issuecomment-1211410066) by @cblecker we need the blessing of SIG Arch to finalize the rename and relocating op OCI-Proxy repo.
It would be good to get that done asap, as reference to register.k8s.io switch-over would be mentioned in the release notes of v1.25 next week.

If we can get a +1 from @johnbelamaric @derekwaynecarr and @dims (this time with your SIG Arch hat on) we can move this forward that would be GREAT. Thanks.

dims:
+1 from me with SIG-Arch hat on :)

johnbelamaric:
+1

dims:
thanks @johnbelamaric

Riaankl:
Thank you @johnbelamaric & @dims 

@nikhita & @mrbobbytables we have the blessing of both SIG K8s Infa and SIG Ach to proceed with the migration and renaming of the OCI-Proxy repo.

mrbobbytables:
repo has been migrated, teams will be added in #3695

BobyMCbobs:
> repo has been migrated, teams will be added in #3695

[...]

PR #3695, opened 2022-09-14
Move oci-proxy teams from kubernetes-sigs to kubernetes org

mrbobbytables:
this removes the sig-k8s-infra dir from kubernetes-sigs, the only teams present were related to the oci-proxy.

ref: #3605


mrbobbytables:
/hold
need to update repo name

PR #3791, opened 2022-10-27
Update admins and maintainers of KMM

qbarrand:
Remove zvonkok and adrianchiris who contributed to SRO but not to KMM.
Set qbarrand and yevgeny-shnaidman who are the main KMM contributors.

### DIFF
--- a/config/kubernetes-sigs/sig-node/teams.yaml
+++ b/config/kubernetes-sigs/sig-node/teams.yaml
@@ -36,17 +36,16 @@ teams:
   kernel-module-management-admins:
     description: Admin access to kernel-module-management repo
     members:
-    - adrianchiris
-    - zvonkok
+    - qbarrand
+    - yevgeny-shnaidman
     privacy: closed
     previously:
     - special-resource-operator-admins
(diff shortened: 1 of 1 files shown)


bart0sh:
@adrianchiris @zvonkok any objections? If not, please lgtm.

adrianchiris:
lgtm, no objections on my end.

mrunalp (approved):

iranzo:
@dchen1107 @derekwaynecarr is everything fine with this change?

derekwaynecarr:
in case I missed a discussion, is there any discussion on this change from a sig meeting or other setting?

SergeyKanzhelev:
once concern I have is that this change make both contributors be from the same company. I don't think this should block the PR, but I strongly suggest to seek for maintainers from other company.

qbarrand:
@derekwaynecarr this was agreed to at Feb 7th sig-node meeting. Names removed were there from historical reasons (SRO -> KMM) but are not KMM contributors. OK to `/unhold`?

PR #3818, opened 2022-11-05
update aojea membership on different orgs

aojea:
- sig-testing 
- sig-network
- ingress-gce

/assign @BenTheElder @thockin @bowei

### DIFF
--- a/config/kubernetes/sig-testing/teams.yaml
+++ b/config/kubernetes/sig-testing/teams.yaml
@@ -5,6 +5,7 @@ teams:
     - cblecker
     members:
     - akutz
+    - aojea
     - BenTheElder
     - bowei
     - chases2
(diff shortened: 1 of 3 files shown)


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -30,6 +31,7 @@ teams:
       sig-testing-leads:
         description: sig-testing chairs and technical leads
         members:
+        - aojea
```
BenTheElder, line 34: Can we add the others as well?

config/kubernetes/sig-testing/teams.yaml
```diff
@@ -39,6 +41,7 @@ teams:
         members:
         - akutz
         - amwat
```
BenTheElder, line 43: This team isn't really in use but the contents are also quite stale, it has some inactive members

config/kubernetes/sig-testing/teams.yaml
aojea: ok, let me close this and send multiple PRs

PR #3869, opened 2022-12-02
Updating Gateway API Admins to match Gateway API Maintainers

robscott:
This mirrors https://github.com/kubernetes-sigs/gateway-api/pull/1528. I'd assumed it's best to keep the groups separate but equal here to match other config, but let me know if I should just merge this into a single group here as well.


thockin:
Thanks!

/lgtm
/approve

PR #3979, opened 2023-01-24
sig-instrumentation: Add mrueg

mrueg:
I'd like to join sig-instrumentation as I'm already maintaining a subproject (kube-state-metrics).

@dgrisonnet


dgrisonnet:
Thanks for the amazing work you are doing on kube-state-metrics :)

/lgtm
/approve

Issue #4073, opened 2023-03-06
Provide a way to automatically add license headers


nikhita:
We periodically see repos being donated to the kubernetes org(s). Donation requires these repos to add boilerplate license headers to all their files. Today, we don't have a way to do this automatically.

https://github.com/kubernetes/repo-infra/pull/82 took a dab at this but was closed due to age.

@dims also pointed a few other alternative approaches:

1. https://github.com/kunalkushwaha/ltag - this is used in containerd
2. https://github.com/google/go-licenses

/help

PR #4077, opened 2023-03-07
Adding automation to add boilerplate license/copyrights headers

mohitsharma-in:
#4073 Adding support to automatically add license headers


nikhita:
/assign

Assigning to myself to approve. Will let @MadhavJivrajani and @Priyankasaggu11929 take an initial review pass at this.

mohitsharma-in:
pull-org-test-all test is failing since `--config` is being passed which is not expected in this test case that would require changes in makefile to run test

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	templatesDir = flag.String("templates", "../boilerplate", "directory containing license templates")
+	excludeDirs  = flag.String("exclude", ".git,vendor,node_modules", "comma-separated list of directories to exclude")
+	path         = flag.String("path", ".", "Absolute Path to directory location")
+	helpFlag     = flag.Bool("help", false, "Display this help message and exit.")
```
Priyankasaggu11929, line 34: nit: remove the fulllstop -> `Display this help message and exit.`

In `excludeDirs`, we may extend the defaults to this list here: https://github.com/kubernetes/org/blob/main/hack/boilerplate/verify_boilerplate.py#L42-L50

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+
+func printHelp(progName string) {
+	fmt.Fprintf(os.Stderr, "Usage: %s [--help] [-d dir] [--exclude ] [--templates]\n", progName)
+	fmt.Fprintln(os.Stderr, "Add a copyright/license header to all source files in a Git repository.\n-")
+	fmt.Fprintln(os.Stderr, "Options:")
+	flag.PrintDefaults()
+	fmt.Fprintln(os.Stderr, "\nExamples:")
+	fmt.Fprintf(os.Stderr, "  %s               # Add license headers to all source files in the current directory and its subdirectories, using the default file extensions and templates directory.\n", progName)
+	fmt.Fprintf(os.Stderr, "  %s --dir /path/to/repo --exclude vendor,node_modules --templates /path/to/templates   # Add license headers to all .go, .sh, .py, makefile, docker files in /path/to/repo and its subdirectories, using the license header templates in /path/to/templates.\n", progName)
+}
```
Priyankasaggu11929, line 85: **Suggestion:** How about using go library like [cobra](https://github.com/spf13/cobra) to implement the command line flags?  It provides a command-line boilerplate and takes care of the common flags like (-h, --help) on its own.

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+	switch {
+	case strings.HasSuffix(path, ".sh"):
+		return true
+	case strings.HasSuffix(path, "Makefile"):
+		return true
+	case strings.HasSuffix(path, "Dockerfile"):
+		return true
+	}
+	return false
+}
```
Priyankasaggu11929, line 105: **suggestion:** instead of using switch statements here to check the file extenstion, how about storing the code/build file extensions as maps and then perform check?

That would improve the search and make it easier to maintain/add new file extensions in future?


```
var codeFileExtensions = map[string]bool{
	".go":    true,
	".c":     true,
	".h":     true,
	...
}

var buildFileExtensions = map[string]bool{
	".sh":       true,
	"Makefile":  true,
	"Dockerfile": true,
}

func isCodeFile(path string) bool {
	return codeFileExts[strings.ToLower(filepath.Ext(path))]
}
...

```

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+
+func hasLicense(path string) bool {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(data), "Copyright")
+}
```
Priyankasaggu11929, line 113: Just checking for the string "Copyright" might produce false positives as there could be cases of use of "Copyright" string in other parts of the file.

Maybe comparing with >1 strings of the known license headers (like with `Licensed under the Apache License` as we have [here in boilerplate](https://github.com/kubernetes/org/tree/main/hack/boilerplate)) and then do the check?


Also, would be helpful to return the error message as well from the `hasLicense()` function.

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+		tmplFilename = "boilerplate.Dockerfile.txt"
+	case filepath.Ext(path) == ".py":
+		tmplFilename = "boilerplate.py.txt"
+	case filepath.Ext(path) == ".go":
+		tmplFilename = "boilerplate.go.txt"
+	default:
+		tmplFilename = "boilerplate.tf.txt"
+	}
+	return tmplFilename
+}
```
Priyankasaggu11929, line 131: **Suggestion:** how about storing the different template files in a struct to do the check. It will help resue/reference the template files in other parts of the program (if required later)?

```
type fileTypeTemplate struct {
    extension string
    template  string
}

var fileTypeTemplates = []fileTypeTemplate{
    {".sh", "boilerplate.sh.txt"},
    {"Makefile", "boilerplate.Makefile.txt"},
     ...
}

func getFileType(path string) string {
    for _, ft := range fileTypeTemplates {
        if strings.HasSuffix(path, ft.extension) || filepath.Base(path) == ft.extension {
            return ft.template
        }
    }
    return "boilerplate.tf.txt"
}
```

hack/k-license/k-license.go
```diff
@@ -0,0 +1,151 @@
...
+
+func addLicense(path, templatesDir, year string) error {
```
Priyankasaggu11929, line 133: Are we handling the exception for generated files - like generated go files? 

Ref: https://github.com/kubernetes/org/blob/main/hack/boilerplate/verify_boilerplate.py#L76-L93

Priyankasaggu11929 (commented):
Thanks for the PR and the contribution @mohitsharma-in. I've left a few comments above.

hack/k-license/k-license.go
mohitsharma-in: Not yet , Thanks for the review i'll get it added

hack/k-license/k-license.go
mohitsharma-in: Will make changes.

hack/k-license/k-license.go
mohitsharma-in: Will it make sense to use everything from struct instead of static templates files. 
since this will be used for generating boilerplates for k,k-sig repos which follows pretty much same templates. 
with this we can avoid --template flag and also creation/copying of templates files.

palnabarun:
Apart from what @Priyankasaggu11929 suggested, `k-license` is a better candidate for the `cmd` directory than `hack`.

hack/k-license/k-license.go
Priyankasaggu11929: Sorry for being not clear above.

> how about storing the different template files in a struct to do the check. 

In the suggestion above, I meant avoiding the use of hardcoded strings for template file names (such as "boilerplate.sh.txt") and extensions (such as ".sh") within the function. Instead, to define these values in a struct format at the top level to make it easier to reference them throughout the code in the future.

To be clear, we will still be using the existing static template files present here -> https://github.com/kubernetes/org/tree/main/hack/boilerplate.

 I have also included some comments in the code suggestion above to provide further clarity.

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,199 @@
...
+
+}
+func (opts *Options) Run() error {
+	excludeList := strings.Split(opts.excludeDirs, ",")
+	files := 0
+	err := filepath.Walk(opts.path, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() && contains(excludeList, info.Name()) {
```
MadhavJivrajani, line 107: I am a little unsure about this. `info.Name()` is going to return only the base name of the dir. In a repo there can be multiple paths with the same base name. I don't think this would work for that.

I can't think of a straightforward way to handle this, but a possible way that could handle more cases like this, but still not all, could be:
Calling
```go
contains(excludeList, path)
```
and `contains` can be something like:
```go
func contains(exclude []string, path string) bool {
  for _, item := range exclude {
    var short, long string
    if len(path) < len(item) {
      short = path
      long = item
    } else {
      // other way around
    }
    if strings.HasSuffix(long, short) { return true }
  }
  return false
}
```
Instead of `HasSuffix`, it might be worth considering `filepath.Rel` too.

cmd/k-license/Readme.md
MadhavJivrajani, line 1: nit: Please change filename to README.md

cmd/k-license/k-license.go
MadhavJivrajani, line 1: I think we should add a `--confirm` flag. If the flag is passed, only then any _actual_ modifications are made by the tool, if not, tool only prints the files that _would_ be modified if the flag is actually passed. 

This will prevent any accidental modifications and also help in cases where multiple unique directory names exist (see comment below).

This will need a unit test.

MadhavJivrajani (changes requested):
Thanks @mohitsharma-in! Leaving a few comments

mohitsharma-in:
@MadhavJivrajani @Priyankasaggu11929 Can you review the PR and provide feedback

cmd/k-license/k-license.go
Priyankasaggu11929: > In a repo there can be multiple paths with the same base name. I don't think this would work for that.

One quick example from say -  kubernetes/test-infra:

Following files are on two different paths in the repo with the same base name:
- https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e/Dockerfile
- https://github.com/kubernetes/test-infra/blob/master/images/builder/Dockerfile

@MadhavJivrajani could you please fill in more (or correct if I understood it wrong). Thank you! :slightly_smiling_face:

Priyankasaggu11929:
@mohitsharma-in, can you please include the suggestion made in https://github.com/kubernetes/org/pull/4077#discussion_r1138585469 to add a new `--confirm` flag to the `k-license` tool to address the failing CI test?

Also, this one (https://github.com/kubernetes/org/pull/4077#discussion_r1132024182) still needs addressing -  to improve the string check for more than just "Copyright" string and return error from the "hasLicense()" function. Can you please review and address this? Thank you.

Once these modifications are made, please mark the implemented review comment threads as `resolved` (just for tracking purposes) and merge the commits into one. I will review it again. Thank you so much for all the changes. :slightly_smiling_face:

hack/k-license/k-license.go
mohitsharma-in: > Also, would be helpful to return the error message as well from the hasLicense() function.
What error message are you expecting from this function, since it only does read file

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,232 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func hasLicense(path string) bool {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return false
```
Priyankasaggu11929, line 179: @mohitsharma-in, answering below for https://github.com/kubernetes/org/pull/4077#discussion_r1165100234

Currently the function is returning a `false` boolean value in case if  ` err != nil`.  Returning the error message along with boolean, would be helpful for debugging.

cmd/k-license/k-license.go
mohitsharma-in: I've checked it test infra repo seems to be working fine

cmd/k-license/k-license.go
mohitsharma-in: flag added

mohitsharma-in:
@MadhavJivrajani  @Priyankasaggu11929  Can you review it again

Priyankasaggu11929:
Thanks so much for the changes @mohitsharma-in. For the failing CI test, let's do the following:

In `Makefile`, update the `test` target as:

```
.PHONY: test
test: config
        CONFIG_PATH=$(MERGED_CONFIG) go test ./...
```

and then in the `config/config_test.go`:

```
func TestMain(m *testing.M) {
        configPath, exists := os.LookupEnv("CONFIG_PATH")
        if ! exists {
                fmt.Println("Environment variable \"CONFIG_PATH\" must be set")
                os.Exit(1)
        }

        raw, err := ioutil.ReadFile(configPath)
        if err != nil {
                fmt.Printf("cannot read generated config.yaml from %s: %v\n", configPath, err)
                os.Exit(1)
        }

        if err := yaml.Unmarshal(raw, &cfg); err != nil {
                fmt.Printf("cannot unmarshal generated config.yaml from %s: %v\n", configPath, err)
                os.Exit(1)
        }

        os.Exit(m.Run())
}

```

config/config_test.go
```diff
@@ -37,20 +37,20 @@ var configPath = flag.String("config", "config.yaml", "Path to generated config"
...
+	if !exists {
+		fmt.Println("Environment variable \"CONFIG_PATH\" must be set")
 		os.Exit(1)
 	}
 
-	raw, err := ioutil.ReadFile(*configPath)
+	raw, err := ioutil.ReadFile(configPath)
 	if err != nil {
-		fmt.Printf("cannot read generated config.yaml from %s: %v\n", *configPath, err)
+		fmt.Printf("cannot read generated config.yaml from %s: %v\n", configPath, err)
```
Priyankasaggu11929, line 48: @mohitsharma-in, nit:

remove the following unused bits of code in `config/config_test.go`

- unused package import - "flag"

- this is no longer required:

```go
var configPath = flag.String("config", "config.yaml", "Path to generated config")
```

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if opts.confirm {
+		fmt.Printf("Modified %v files\n", files)
+	} else {
+		fmt.Printf("Skipping modification. Dry running \t")
+		fmt.Printf("%v files will be modified\n", files)
```
Priyankasaggu11929, line 142: @mohitsharma-in, 

```
Skipping modification. Dry running 	1 files will be modified
```

In the above message, along with tracking the **count of files to be modified**, could we **also track & output their names (path)** as well? 

That would make above message more useful to understand where actual changes will be made.

Also, in case `--confirm` flag is **not set:**

let's print upfront (something like):

```
DRY RUN: No file changes will be made! To make file modifications, rerun the command with  "--confirm" flag
```

And then, if:
- (1) if **count(files) == 0,** just print something like - "All files have appropriate License Headers. No changes required."
- (2) if **count(files) >0**, let's print something like - "N files will be modified to add License Headers" and then list their names!

Priyankasaggu11929 (changes requested):

cmd/k-license/k-license.go
mohitsharma-in: ack

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil
+	})
+	if opts.confirm {
+		fmt.Printf("Modified %v files\n", files)
+	} else {
+		fmt.Printf("DRY RUN: No file changes will be made! To make file modifications, rerun the command with  \"--confirm\" flag\n")
+		if files == 0 {
+			fmt.Printf("All files have appropriate License Headers. No changes required.\n")
+		}
+		if files > 1 {
```
Priyankasaggu11929, line 147:
```suggestion
		if files > 0 {
```

@mohitsharma-in ^

Priyankasaggu11929 (changes requested):
@mohitsharma-in, all looks look fine to me. Thanks so much for the contribution!

@nikhita @MadhavJivrajani could you please do a final review for lgtm/approval.

cmd/k-license/k-license.go
mohitsharma-in: ack

mohitsharma-in:
> Thanks @mohitsharma-in! Leaving a few comments

resolved

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,259 @@
...
+	addCmd.Flags().StringVar(&opts.path, "path", ".", "Defaults to Current directory")
+	addCmd.Flags().BoolVar(&opts.confirm, "confirm", false, "")
+	rootCmd.AddCommand(addCmd)
+	if err := rootCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+
+}
+func (opts *Options) Run() error {
+	//excludeList := strings.Split(opts.excludeDirs, ",")
```
MadhavJivrajani, line 111: Should we remove this?

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,259 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			return opts.Run()
+		},
+	}
+
+	addCmd.Flags().StringVar(&opts.templatesDir, "templates", "../../hack/boilerplate", "directory containing license templates")
+	addCmd.Flags().StringSliceVarP(&opts.excludeDirs, "exclude", "e", excludeDirsLocations, "comma-separated list of directories to exclude")
+	addCmd.Flags().StringVar(&opts.path, "path", ".", "Defaults to Current directory")
+	addCmd.Flags().BoolVar(&opts.confirm, "confirm", false, "")
```
MadhavJivrajani, line 103:
```suggestion
	addCmd.Flags().BoolVar(&opts.confirm, "confirm", false, "confirm actually adding license boilerplate to files")
```

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,259 @@
...
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "k-license",
+		Short: "Tool for Adding license Headers",
+	}
+	addCmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add Headers to files",
+		RunE: func(cmd *cobra.Command, args []string) error {
+
```
MadhavJivrajani, line 95:
```suggestion
```

cmd/k-license/k-license_test.go
```diff
@@ -0,0 +1,32 @@
...
+limitations under the License.
+*/
+
+package main
+
+import (
+	"testing"
+)
+
+var opt_test = &Options{}
```
MadhavJivrajani, line 23:
```suggestion
var optTest = &Options{}
```
nit: Go coding convention

cmd/k-license/k-license_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+// Running E2E Test
+func TestAddLicense(t *testing.T) {
+	opt_test.templatesDir = "../../hack/boilerplate"
+	opt_test.excludeDirs = excludeDirsLocations
+	opt_test.path = "../"
+	opt_test.confirm = false
+	opt_test.Run()
```
MadhavJivrajani, line 31: 2 things:
- Can we confirm that no changes have been made? 
  - We can have a `./k-license/testdata` dir and have a simple Go file there for now without a licence header.
  - One test case: set `confirm`, verify changes happen
  - One test case: set `confirm`, verify no changes happen
  - [Not needed in this PR, but maybe a followup] the `testdata` can have some excluded files too and we can test for that.
- We need to check for error returned by `Run()` and fail the test if there is an error.

MadhavJivrajani (commented):
Sorry @mohitsharma-in for not taking a look, this dropped off my radar.
The implementation by and large looks good, I just have a few minor suggestions to make, LGTM otherwise. Please ping me once done, and I will approve it.

Thank you again for your contribution, and apologies for the delay.

cmd/k-license/k-license.go
```diff
@@ -0,0 +1,259 @@
...
+			fmt.Printf("Listing files to be modified:\n")
+			for _, file := range fileList {
+				fmt.Printf("%s\n", file)
+			}
+		}
+
+	}
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
MadhavJivrajani, line 158: Let's return the error here, `cobra` will exit with a non 0 exit code if an error is returned to `RunE`.

mohitsharma-in:
@MadhavJivrajani  Now the existing boilerplate verify is failing due to test data and reccomendations ?

cmd/k-license/k-license_test.go
```diff
@@ -0,0 +1,40 @@
...
+	optTest.Run()
+}
+
+func TestAddConfirmLicense(t *testing.T) {
+	optTest.templatesDir = "../../hack/boilerplate"
+	optTest.excludeDirs = excludeDirsLocations
+	optTest.path = "./testdata"
+	optTest.confirm = true
+	optTest.Run()
+}
```
MadhavJivrajani, line 40: This still isn't testing what we want:
1. Need to verify if err returned is nil or not
2. Need to confirm if the files actually change or not based on `confirm`.
Something like:
```go
func TestAddLicense(t *testing.T) {
	cases := []struct {
		confirm bool
		desc    string
	}{
		{
			confirm: false,
			desc:    "confirm=true, files should actually change",
		},
		{
			confirm: true,
			desc:    "confirm=false, files should not actually change",
		},
	}
	for _, c := range cases {
		t.Run(c.desc, func(t *testing.T) {
			optTest.templatesDir = "../../hack/boilerplate"
			optTest.excludeDirs = excludeDirsLocations
			optTest.confirm = c.confirm
			// create a tmp dir with the test data that we can delete and
			// recreate after each test case.
			tmpTestDir := createTmpDir()
			optTest.path = tmpTestDir
			if err := optTest.Run(); err != nil {
				t.Errorf("error running test case %s: %v", c.desc, err)
			}
		})
	}
}
```

mohitsharma-in:
@MadhavJivrajani  Can you please review it.

mohitsharma-in:
@MadhavJivrajani  Please review it once again

MadhavJivrajani (commented):
Just 2 small comments, otherwise LGTM.
Thank you so much for persisting through reviews here, and I apologise this has taken so long.

justaugustus (changes requested):
Please strongly consider using `addlicense` instead: https://github.com/google/addlicense

There may also be some useful content here:
- https://github.com/kubernetes-sigs/release-utils/blob/main/mage/boilerplate.go
- https://github.com/kubernetes/repo-infra/

mohitsharma-in:
@MadhavJivrajani / @Priyankasaggu11929  
Do you think we should not invest much time on this but rather setup our jobs to use
addlicense instead: https://github.com/google/addlicense as suggested by @justaugustus  above 
https://github.com/kubernetes/org/pull/4077#pullrequestreview-1611085768

Priyankasaggu11929:
Hey @mohitsharma-in, just checked out the discussion on https://github.com/kubernetes/repo-infra/pull/82.

It seems like we do originally wanted to use [verify-boilerplate.py](https://github.com/kubernetes/repo-infra/blob/master/hack/verify_boilerplate.py) from `k/repo-infra`, but back then, we were looking for examples of how to use it and vendor it efficiently into other repositories.

Now, we have a working example in `k-sigs/release-utils` ([boilerplate.go](https://github.com/kubernetes-sigs/release-utils/blob/main/mage/boilerplate.go)), which uses the above script to achieve what we need.

So, yea, I'm +1 on reusing this existing tooling to avoid duplication.

mohitsharma-in:
@Priyankasaggu11929  / @MadhavJivrajani  Please suggest the appropriate here. If we Don't need this one I'll mark it close.

Priyankasaggu11929:
@mohitsharma-in, thank you *so much* for all your work on this PR.

This discussion resurfaced a lot of useful resources!

PR #4117, opened 2023-03-23
sig-net: transfer co-chair from dcbw to MikeZappa87

dcbw:
Per chair transition procedure, there were no objections to Mike being proposed as dcbw's replacement, thus Mike is now a chair.

https://groups.google.com/g/kubernetes-sig-network/c/sZF25bnqZ-8

@MikeZappa87 @thockin @shaneutt


shaneutt (approved):

thockin (commented):
Thanks!

/lgtm
/approve

PR #4153, opened 2023-04-10
Add new leads to sig-contribex teams

nikhita:
Ref: https://github.com/kubernetes/community/issues/7246

/assign @mrbobbytables


mrbobbytables:
/lgtm
oof those teams are REALLY outdated and should be cleaned up 😬

PR #4208, opened 2023-05-11
Add msau42/deads2k to api-approvers/api-reviewers lists

liggitt:
Sync with api-approvers from https://github.com/kubernetes/kubernetes/blob/e1ad9bee5bba8fbe85a6bf6201379ce8b1a611b1/OWNERS_ALIASES#L318-L324


palnabarun:
/approve
(as an org admin)

PR #4211, opened 2023-05-11
Add mboersma to image-builder-admins

mboersma:
Syncs up the GH ownership with image-builder's [`OWNERS_ALIASES`](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/OWNERS_ALIASES#L18)

### DIFF
--- a/config/kubernetes-sigs/sig-cluster-lifecycle/teams.yaml
+++ b/config/kubernetes-sigs/sig-cluster-lifecycle/teams.yaml
@@ -301,25 +301,16 @@ teams:
   image-builder-admins:
     description: admin access to image-builder
     members:
-    - CecileRobertMichon
-    - codenrhoden
     - jsturtevant
-    - justinsb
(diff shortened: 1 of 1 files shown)


config/kubernetes-sigs/sig-cluster-lifecycle/teams.yaml
```diff
@@ -307,6 +307,7 @@ teams:
     - justinsb
     - kkeshavamurthy
     - luxas
+    - mboersma
```
CecileRobertMichon, line 310: could you please update the whole list? I see several names (include myself) that shouldn't be in here anymore

mboersma: Updated so both lists are just the three current maintainers.

CecileRobertMichon (commented):
/lgtm
/approve

PR #4274, opened 2023-06-08
Rotate SIG Testing chairs

michelle192837:
Ref https://github.com/kubernetes/community/issues/7346

Also updated membership a bit more, since some lists look inconsistent. (Some folks who are no longer involved in SIG Testing should probably also be removed but I want to leave that for later.)


mrbobbytables (commented):
woops, this PR got missed :x

/lgtm
/approve

Issue #3094, opened 2021-11-12
Cleanup Inactive Org Members


palnabarun:
Following is the list of [inactive members](https://github.com/kubernetes/community/blob/master/community-membership.md#inactive-members) who are slated to be removed from the Kubernetes organizations: 

**TBD** (pending on a data import from devstats)

### Tasks
- [ ] Get the data import
- [ ] Find members to remove and teams they belong to
- [ ] Ping and wait for people to let us know if they have been contributing to non-code areas
- [ ] Remove the inactive members

palnabarun:
Previous Instances of cleanup: https://github.com/kubernetes/org/issues/2456, https://github.com/kubernetes/org/issues/2076

mrbobbytables:
/assign
(just to keep it on my radar)

mrbobbytables:
/retitle Cleanup Inactive Org Members -Q1 2022

nikhita:
/unassign @nikhita @MadhavJivrajani @RinkiyaKeDad 

Let's re-assign when tasks are being picked up.

@palnabarun do you know how to import the required data from devstats? If yes, can you also help document this so either of us can pick this up in the future?

Sea-n:
We still need to cleanup inactive org members.
/remove-lifecycle stale

It's being postponing, remove the time from issue title.
/retitle Cleanup Inactive Org Members

nikhita:
/remove-lifecycle frozen

@palnabarun or @mrbobbytables we've added a lot of members in the last few months. Do y'all have bandwidth to cleanup inactive members soon?

pacoxu:
/assign 
I could like to help once we need to do a cleanup.

palnabarun:
PR #4291  introduces a command to audit the orgs. There are few TODOs to be done as followups after it is merged, most important being removing org members using the `korg` toolkit. I will take that up next week to ease the removal process.

pacoxu:
It seems that we did not cleanup inactive org members for two years. 

Can I try with @palnabarun's tool in the PR to do a cleanup this month?

pacoxu:
The removal list is at https://gist.github.com/pacoxu/21622b30d3511d8a9df8507f58926769. 

I would open a new issue  "Clean up inactive members - Sept 2023" to track the removal process if there is no question about it.

palnabarun:
@pacoxu I will send comms about the removal next week after verifying the audit script. We need to give the community a heads-up before removing org members.

[...]

PR #4291, opened 2023-06-14
korg: add audit command

palnabarun:
Implements the audit command

### Top level logic

1. Fetch all users from the org configuration
2. Fetches contribution statistics from [devstats](https://k8s.devstats.cncf.io/d/13/developer-activity-counts-by-repository-group?orgId=1&var-period_name=Last%20year&var-metric=contributions&var-repogroup_name=All&editPanel=4&var-repo_name=kubernetes%2Fkubernetes&var-country_name=All)
3. Parses an exception file (in CSV format), if specified
4. Filters out members with contributions exceeding a threshold or present in the exceptions file. Note: The filtered-out members are active contributors, and inactive members are the ones left.
5. Checks if the users are owners.
6. Print the data to a Markdown file or to stdout.

### Example

Running the following with [this exceptions file](https://gist.github.com/palnabarun/ea531edf42c76d0d1135561843e7b6b0#file-exceptions-csv) after checking out this branch will give [this table of users](https://gist.github.com/palnabarun/ea531edf42c76d0d1135561843e7b6b0#file-removals-md) to be removed.
```
go run ./cmd/korg/... audit --org kubernetes,kubernetes-sigs,kubernetes-client,kubernetes-csi --activity-threshold 0 --output-file removals.md --exceptions-file exceptions.csv --check-teams --check-owners
```

### Follow-ups

- Post the data to https://github.com/kubernetes/org/issues/3094
- Implement the remove command


nikhita:
/assign
for final review and lgtm

/assign @MadhavJivrajani @Priyankasaggu11929 
can you take an initial pass?

cmd/korg/audit.go
```diff
@@ -0,0 +1,336 @@
...
+    and period = '%s'
+  group by
+    split_part(name, '$$$', 1)
+) sub`, period),
+				Format: "table",
+			},
+		},
+	}
+
+	requestBody, _ := json.Marshal(postBody)
```
MadhavJivrajani, line 127: Is there a reason we are ignoring the error here?

cmd/korg/korg.go
```diff
@@ -146,9 +150,35 @@ func main() {
...
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return OrgAudit(o)
+		},
+	}
+
+	// flags
+	auditCmd.Flags().IntVar(&o.ActivityThreshold, "activity-threshold", 50, "minimum activity to be considered active. default: 50")
```
MadhavJivrajani, line 172: Can we add validation for non-negativeness of `activity-threshold` in `PreRunE`? Or is that a valid input? If negative is passed, all users would be above threshold I think.

cmd/korg/audit.go
MadhavJivrajani, line 1: This is probably something for the future, and thinking out loud here: would it make sense to have "if a username is still a valid GH username or not?" as a default filtering predicate? If they are not, then irrespective or threshold and contrib count, we suggest this as a candidate for pruning.

cmd/korg/korg.go
```diff
@@ -146,9 +150,35 @@ func main() {
...
+	auditCmd.Flags().StringVar(&o.Period, "period", "y", "period to look back for activity. default: y (Year)")
```
MadhavJivrajani, line 173: What other values can this take according to devstats and in what format? It might be worth documenting it in the flag desc.

MadhavJivrajani (commented):
Just a few minor suggestions and questions.
Overall, looks great! Thank you @palnabarun, this is going to be very handy ❤️

cmd/korg/utils.go
```diff
@@ -121,3 +126,34 @@ func caseAgnosticSort(arr []string) {
...
+
+	err = json.Unmarshal(body, &r)
+	if err != nil {
+		return false, err
+	}
+
+	found := false
+	for _, matches := range r.Results {
+		if len(matches.Matches) != 0 {
+			found = true
```
pacoxu, line 154: We can add `break` here. 
Or remove the found bool, and `return true, nil` here and `return false, nil` in the end.

Or another way is returning`r.Status.FilesOpened > 0`.

cmd/korg/audit.go
```diff
@@ -0,0 +1,336 @@
...
+		table.SetHeader([]string{"Username", "Orgs", "Owner?", "Owners Link"})
+
+		for _, v := range orgMembersBelowThresholdAfterException {
+
+			table.Append([]string{
+				v.Username,
+				strings.Join(v.Orgs, ", "),
+				fmt.Sprintf("%v", v.IsOwner),
+				fmt.Sprintf("https://go.k8s.io/owners/%s", v.Username),
+			})
```
Priyankasaggu11929, line 302: @palnabarun, thanks a lot! It looks great to me.

Just a small suggestion — Can we render the owner URL only when the username _is_ actually an owner? It's just to make the owner usernames stand out visually in the table. 

That's all. Thank you! :slightly_smiling_face:  

```suggestion
                        row := []string{
                                v.Username,
                                strings.Join(v.Orgs, ", "),
                                fmt.Sprintf("%v", v.IsOwner),
                        }

                        if v.IsOwner {
                                row = append(row, fmt.Sprintf("https://go.k8s.io/owners/%s", v.Username))
                        } else {
                                row = append(row, "") // Add an empty string as a placeholder
                        }

                        table.Append(row)
```

Priyankasaggu11929:
Also — it would be useful to include [this example exceptions file](https://gist.github.com/palnabarun/fbce39dea32198871fc2c586c20194f5#file-exceptions-csv) as well in the PR.

Both for format reference and maybe as a default entries for bot accounts?

(`k8s-release-robot` is appearing in the output table for usernames to remove. Perhaps we could add that to the exceptions list too. )

pacoxu:
There is a special case that not sure if we want to make a flag.

https://go.k8s.io/owners/liu-cong is in the owner file but is emeritus. For such cases, we don't need an owner cleanup.
For others, he/she may be in the owner file and need a cleanup.

palnabarun:
> [...] would it make sense to have "if a username is still a valid GH username or not?" as a default filtering predicate? [...]

@MadhavJivrajani The idea is good. I think it is taken care of already. Peribolos occasionally fails to reconcile if an invalid user exists in our manifests. Wondering if that is a sufficient lazy check.

--- 

@Priyankasaggu11929 @MadhavJivrajani @pacoxu I resolved all your pending comments and regenerated the lists.

Functionality diff from previous version:
- Removed the output type. Rendering to the terminal doesn't provide much value.
- The script now computes all teams that the users are part of.
- Takes care of case agnostic behavior of GitHub usernames. The repo has entries for the same user in correct case and lower case.

docs/sample-exceptions.csv
```diff
@@ -0,0 +1,10 @@
+username,reason
+k8s-publishing-bot,bot account
+k8s-infra-ci-robot,bot account
+k8s-github-robot,bot account
+k8s-ci-robot,bot account
+k8s-ci-robot,bot account
+k8s-publishing-bot,bot account
+k8s-release-bot,bot account
```
Priyankasaggu11929, line 8:
```suggestion
k8s-release-bot,bot account
k8s-release-robot,bot account
```

docs/sample-exceptions.csv
```diff
@@ -0,0 +1,10 @@
+username,reason
+k8s-publishing-bot,bot account
+k8s-infra-ci-robot,bot account
+k8s-github-robot,bot account
+k8s-ci-robot,bot account
+k8s-ci-robot,bot account
```
Priyankasaggu11929, line 6:
```suggestion
k8s-ci-robot,bot account
```

cmd/korg/audit.go
```diff
@@ -0,0 +1,375 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, org := range orgs {
+		for _, u := range append(config[org].Members, config[org].Admins...) {
+			if user, found := users[strings.ToLower(u)]; found {
+				if user.Orgs == nil {
+					user.Orgs = []string{}
+				}
```
MadhavJivrajani, line 95: Do we need this branch? Or is it just in there as a sanity check? Because we'll never really hit this iiuc.

cmd/korg/audit.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+		}
+
+		for teamName, team := range config[org].Teams {
+			for _, u := range append(team.Members, team.Maintainers...) {
+				if user, found := users[strings.ToLower(u)]; found {
+					if user.Teams == nil {
+						user.Teams = make(map[string][]string)
+					}
```
MadhavJivrajani, line 111: Same comment as above

cmd/korg/audit.go
```diff
@@ -0,0 +1,375 @@
...
+
+	return exceptions, nil
+}
+
+func usernameNotInContributors(contribs map[string]Contribution, username string) bool {
+	if _, found := contribs[username]; !found {
+		return true
+	}
+
+	return false
```
MadhavJivrajani, line 222:
```suggestion
	_, found := contribs[username]
	return !found
```

MadhavJivrajani (commented):
A few non-blocking comments, LGTM otherwise!

MadhavJivrajani:
> The idea is good. I think it is taken care of already. Peribolos occasionally fails to reconcile if an invalid user exists in our manifests. Wondering if that is a sufficient lazy check.

I think this should suffice, thanks for reminding me about that 👍🏼

cmd/korg/audit.go
palnabarun: this is just a sanity check.

cmd/korg/audit.go
palnabarun: this is required to make sure assignment is not made to a nil map.

MadhavJivrajani (approved):
/lgtm

Priyankasaggu11929 (approved):

pacoxu:
Good job. 

My last run two months ago resulted in a long list.

PR #4321, opened 2023-07-06
Offboard Luke Hinds from SRC

cji:
Luke has stepped down from the SRC (https://github.com/kubernetes/committee-security-response/issues/182) and this PR is part of our offboarding process.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -165,7 +165,6 @@ aliases:
     - cji
     - enj
     - joelsmith
-    - lukehinds
     - micahhausler
     - ritazh
     - tabbysable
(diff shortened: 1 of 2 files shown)


mrbobbytables (commented):
There's one other place that needs updating in k/org - the root OWNERS_ALIASES file

https://github.com/kubernetes/org/blob/d85a7357e3aef21478bebaccd4d660a633e691fd/OWNERS_ALIASES#L162-L171

cji:
Thanks @mrbobbytables! I removed Luke from that file now as well.

mrbobbytables (commented):
Thanks!
/lgtm
/approve

PR #4384, opened 2023-08-14
Update CAPO maintainers

mdbooth:
We don't differentiate between maintainers admins and maintainers, so these lists are intended to be identical.


neolit123 (commented):
/lgtm
/approve

PR #4390, opened 2023-08-15
Add dims to test-infra-admins + test-infra-maintainers

dims:
Most of the folks on these 2 teams are no longer active. Putting my hands up to help with the repo. (Triggered right now by the need for adding `/lgtm`'s to some PRs that hit a github webhook hiccup!)

### DIFF
--- a/config/kubernetes/sig-testing/teams.yaml
+++ b/config/kubernetes/sig-testing/teams.yaml
@@ -66,6 +66,7 @@ teams:
     - chaodaiG
     - chases2
     - cjwagner
+    - dims
     - fejta
     - k8s-infra-ci-robot
     - listx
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -66,6 +66,7 @@ teams:
     - chaodaiG
     - chases2
     - cjwagner
+    - dims
```
BenTheElder, line 69: we can do this in a follow up but let's review this team, we've reviewed these in OWNERS before but clearly need to go back over them here

BenTheElder: I would also suggest, perhaps in another PR, that we should add @aojea and @pohly, our other SIG TLs

dims: Sounds good. let me close this for now

Issue #4365, opened 2023-08-02
[Tracking Issue - Peribolos Incident] Peribolos removing repos and members from under teams


MadhavJivrajani:
Recently we added the ability to manage team repo permissions using Peribolos: https://github.com/kubernetes/org/pull/2614
However, as part of this change, the team configurations that went in ended up being skewed with what the desired state of the world should have been.

Peribolos looks at team configuration and reconciles that state with GitHub. Due to faulty config as part of the above change, it started to remove members and repos out from teams. The most severe example of this being the publishing bot failing: https://github.com/kubernetes/kubernetes/issues/56876#issuecomment-1660608409

The mitigation strategy being followed to bring things back to the desired state is as follows:

- [x] Stop peribolos from syning state to prevent more such acts of removal
  - https://github.com/kubernetes/test-infra/pull/30263
- [x] Add functionality to Peribolos to allow skipping any type of removals (this will help us re-enable peribolos to sync back the state that was removed while not removing more till we ensure the configs that peribolos reads is accurate desired state of the world
  - https://github.com/kubernetes/test-infra/pull/30264
  - https://github.com/kubernetes/test-infra/pull/30273
- [x] Add back the state that was removed by peribolos to the config files as a way to rectify what was done
  - https://github.com/kubernetes/org/pull/4363
  - https://github.com/kubernetes/org/pull/4364
- [x] Re-enable syncing in Peribolos jobs
  - https://github.com/kubernetes/test-infra/pull/30266
- [x] If things work as expected, wait for a few successful runs and manually verify the config that exists currently
  - [x] https://github.com/kubernetes/org/pull/4370
  - [x] https://github.com/kubernetes/org/pull/4374
- [x] If config is good, remove `--skip-removals` from `./admin/update.sh` after 1-2 successful runs.
  -  https://github.com/kubernetes/org/pull/4392
- [x] Finally remove the `--skip-removals` functionality from Peribolos altogether.
  - https://github.com/kubernetes/test-infra/pull/30392

/assign @MadhavJivrajani @Priyankasaggu11929 
/sig contributor-experience
/area github/management

Priyankasaggu11929:
Another item for investigation (from https://github.com/kubernetes/org/pull/4370#issuecomment-1673144988):

---

**Important Note!**

For some reason, this time `make verify` didn't flag the absence of `design-proposals-archive` and `registry.k8s.io` repo entries in `config/restrictions.yaml`, even before I whitelisted them.
[...]

PR #4392, opened 2023-08-16
Revert "update admin/update.sh to include --skip-removals flag"

MadhavJivrajani:
Now that peribolos is back on its feet, we don't need the skip-removals flag.

Ref https://github.com/kubernetes/org/issues/4365

/assign @Priyankasaggu11929 
/area github/management
/sig contributor-experience

### DIFF
--- a/admin/update.sh
+++ b/admin/update.sh
@@ -40,7 +40,6 @@ args=(
   --fix-teams
   --fix-team-members
   --fix-team-repos
-  --skip-removals
   "${admins[@]/#/--required-admins=}"
 )
 


Priyankasaggu11929:
/lgtm
/approve

Thanks so much! 

@MadhavJivrajani, just confirming —   Once this PR is merged, the plan is to have a few successful runs of Peribolos jobs and then remove the `--skip-removals` flag functionality from test-infra as well?

MadhavJivrajani:
> Once this PR is merged, the plan is to have a few successful runs of Peribolos jobs and then remove the --skip-removals flag functionality from test-infra as well?

You're right, let's wait before getting the test-infra revert in!

PR #4403, opened 2023-08-21
Removing marosset from release-engineering team

marosset:
Part of https://github.com/kubernetes/sig-release/issues/2303

/assign @jeremyrickard


marosset:
/close

I'm part of a Branch Manager Shadow for v1.29 and will re-open this PR after the v1.29 release cycle

PR #4404, opened 2023-08-22
Update test-infra repo admins and maintainers

dims:
active folks from: https://github.com/kubernetes/test-infra/blob/master/OWNERS_ALIASES#L125-L133

### DIFF
--- a/config/kubernetes/sig-testing/teams.yaml
+++ b/config/kubernetes/sig-testing/teams.yaml
@@ -62,16 +62,13 @@ teams:
   test-infra-admins:
     description: admin access to test-infra
     members:
+    - alvaroaleman
+    - aojea
     - BenTheElder
-    - chaodaiG
-    - chases2
     - cjwagner
-    - fejta
     - k8s-infra-ci-robot
-    - listx
     - michelle192837
-    - mpherman2
-    - spiffxp
+    - pohly
     - stevekuznetsov
     privacy: closed
     repos:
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-testing/teams.yaml
```diff
@@ -62,16 +62,13 @@ teams:
   test-infra-admins:
     description: admin access to test-infra
     members:
+    - alvaroaleman
+    - aojea
     - BenTheElder
-    - chaodaiG
-    - chases2
```
michelle192837, line 67: I'd like to leave the folks who are still oncall (chases2, listx, mpherman2)

But aside from that, LGTM!

michelle192837 (approved):
/hold

For others to review

config/kubernetes/sig-testing/teams.yaml
dims: @michelle192837 then shouldn't we also have the `sig-k8s-infra-oncall` as well?

https://github.com/kubernetes/test-infra/blob/master/OWNERS_ALIASES#L16C3-L22

michelle192837: That's fine by me 👍

dims:
it's been a week, let's merge and iterate if needed.

/hold cancel

dims:
will file a follow up PR with the rest of the crew once this merges

PR #4421, opened 2023-08-30
cmd/merge: use strict unmarshalling for merging configs

MadhavJivrajani:
### Commit 1
Strict unmarshalling allows us to fail when we see an unrecognised
key. This helps prevent subtle issues like `repo` being written instead
of `repos`. Previously, the unmarshalling would have ignored this. This
is quite serious because now this means that the desired state of a team
is that there is no repo under it when there clearly is a repo mentioned.
Hence, peribolos would remove this repo.

Strict unmarshalling prevents us from doing that.

Plus, this uses the yaml packages maintained by the Kubernetes community,
and this is also used in test-infra.

### Commit 2

There are now more than 1 test files in the repo, using
`./...` with a flag will not suffice if that flag is not
defined in those test files.

---

One such issue that could've been prevented by strict unmarshalling: https://kubernetes.slack.com/archives/C01672LSZL0/p1692808314288179 

/assign @Priyankasaggu11929 @palnabarun @nikhita

### DIFF
--- a/cmd/merge/main.go
+++ b/cmd/merge/main.go
@@ -26,8 +26,8 @@ import (
 
 	"k8s.io/test-infra/prow/config/org"
 
-	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/yaml"
 )
 
 func parseKeyValue(s string) (string, string) {
@@ -97,13 +97,18 @@ func main() {
 	fmt.Println(string(out))
 }
 
-func unmarshal(path string) (*org.Config, error) {
+func unmarshalFromFile(path string) (*org.Config, error) {
 	buf, err := ioutil.ReadFile(path)
 	if err != nil {
 		return nil, fmt.Errorf("read: %v", err)
 	}
+
+	return unmarshal(buf)
+}
+
+func unmarshal(buf []byte) (*org.Config, error) {
 	var cfg org.Config
-	if err := yaml.Unmarshal(buf, &cfg); err != nil {
+	if err := yaml.Unmarshal(buf, &cfg, yaml.DisallowUnknownFields); err != nil {
 		return nil, fmt.Errorf("unmarshal: %v", err)
 	}
 	return &cfg, nil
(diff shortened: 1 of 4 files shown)


cmd/merge/main.go
```diff
@@ -97,13 +97,18 @@ func main() {
...
 		return nil, fmt.Errorf("read: %v", err)
 	}
+
+	return unmarshal(buf)
+}
+
+func unmarshal(buf []byte) (*org.Config, error) {
 	var cfg org.Config
-	if err := yaml.Unmarshal(buf, &cfg); err != nil {
+	if err := yaml.Unmarshal(buf, &cfg, yaml.DisallowUnknownFields); err != nil {
```
Priyankasaggu11929, line 111: Just read about the [DisallowUnknownFields](https://github.com/kubernetes-sigs/yaml/blob/65d71bb2ae7234d12629f43f77e1a44512057bba/yaml_go110.go#L26) . Yes, I think it's doing exactly what we needed (i.e., to flag unknown fields, rather than dropping them altogether) :+1: 

Tested it. Works beautifully!

```
FATA[0000] Failed to load orgs: merge teams config/kubernetes/org.yaml: error in config/kubernetes/sig-cli/teams.yaml: unmarshal: error unmarshaling JSON: while decoding JSON: json: unknown field "repo" 
```

config/config_test.go
```diff
@@ -32,25 +31,23 @@ import (
...
-		fmt.Printf("cannot read generated config.yaml from %s: %v\n", *configPath, err)
+		fmt.Printf("cannot read generated config.yaml from %s: %v\n", configPath, err)
 		os.Exit(1)
 	}
 
 	if err := yaml.Unmarshal(raw, &cfg); err != nil {
-		fmt.Printf("cannot unmarshal generated config.yaml from %s: %v\n", *configPath, err)
+		fmt.Printf("cannot unmarshal generated config.yaml from %s: %v\n", configPath, err)
 		os.Exit(1)
 	}
```
Priyankasaggu11929, line 52: +1 to above changes.

Priyankasaggu11929 (commented):
@MadhavJivrajani, thanks for the PR.

The changes looks fine to me, and works as I expected.

/lgtm

Will let other folks also chime in!

nikhita (approved):
/lgtm
/approve

palnabarun (approved):
/lgtm
/approve

MadhavJivrajani:
Thanks y'all!
/hold cancel

PR #4435, opened 2023-09-02
Migrating away from deprecated ioutil

Atharva-Shinde:
This PR removes the use of ioutil package as it has been deprecated since Go 1.16


palnabarun (approved):
/lgtm

PR #4463, opened 2023-09-14
Add sig-scheduling-approvers and cleanup teams

sanposhiho:
Ref: https://github.com/kubernetes/kubernetes/pull/119396#issuecomment-1717022820

---

Sometimes, we want to ping approvers only especially when we want to ask for `/approve`.
I always do that with @ kubernetes/sig-scheduling-leads + @ kerthcet ([example](https://github.com/kubernetes/kubernetes/pull/119396#pullrequestreview-1623468480)), it's easier to have a group for approvers for such cases.

---

Also, we have several teams with the same members. This PR also removes them, and only keeps sig-scheduling-leads, sig-scheduling-approvers and sig-scheduling-misc.

### DIFF
--- a/config/kubernetes/sig-scheduling/teams.yaml
+++ b/config/kubernetes/sig-scheduling/teams.yaml
@@ -1,56 +1,8 @@
 teams:
-  sig-scheduling-api-reviews:
-    description: ""
-    members:
-    - ahg-g
-    - alculquicondor
-    - chendave
-    - damemi
-    - denkensk
-    - Huang-Wei
-    - kerthcet
-    - sanposhiho
-    privacy: closed
-  sig-scheduling-bugs:
-    description: ""
-    members:
-    - ahg-g
-    - alculquicondor
-    - chendave
-    - damemi
-    - denkensk
-    - Huang-Wei
-    - kerthcet
-    - sanposhiho
-    privacy: closed
-  sig-scheduling-feature-requests:
-    description: ""
-    members:
-    - ahg-g
-    - alculquicondor
-    - chendave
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-scheduling/teams.yaml
```diff
@@ -61,6 +61,15 @@ teams:
     - kerthcet
     - sanposhiho
     privacy: closed
+  sig-scheduling-approvers:
```
kerthcet, line 64: For discussion: Can we only keep three teams for clearness, like `sig-scheduling-leads`, `sig-scheduling-approvers` and `sig-scheduling-misc` since most of the team members are the same.

sanposhiho: +1 unless other teams are created by a rule in the community (since other sigs also have xxx-bugs, xxx-misc etc, there _may_ be some rules about teams? not sure.)

Huang-Wei: > For discussion: Can we only keep three teams for clearness, like `sig-scheduling-leads`, `sig-scheduling-approvers` and `sig-scheduling-misc` since most of the team members are the same.

SGTM. @sanposhiho could you update this PR by de-duping redundant aliases? (raising a separate one is also fine, but maybe not worth it)

sanposhiho: 👍 Done.

kerthcet: /lgtm

PR #4585, opened 2023-11-14
Add prow owners as members

timwangmusic:
We are adding Prow source owners as members of the @kubernetes-sigs org. This would resolve [issue #68](https://github.com/kubernetes-sigs/prow/issues/68)

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -56,6 +56,7 @@ members:
 - alexander-demicev
 - alexeldeib
 - alexgervais
+- AlexNPavel
 - alexzielenski
 - alijahnas
 - alisondy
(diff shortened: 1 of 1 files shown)


config/kubernetes-sigs/org.yaml
```diff
@@ -476,6 +478,7 @@ members:
 - jlamillan
 - jlsong01
 - jmhbnz
+- jmguzik
```
Priyankasaggu11929, line 481: need alpha sorting here to fix the failing job

```suggestion
- jmguzik
- jmhbnz
```

Priyankasaggu11929 (changes requested):

config/kubernetes-sigs/org.yaml
Priyankasaggu11929: commit - 4ad04e37e8c23eceeb878f23b4d23d2a92b064c2

timwangmusic: Fixed, thanks for pointing out this.

Priyankasaggu11929:
Thank you, @timwangmusic

/lgtm
/approve

PR #4588, opened 2023-11-16
Add cartermckinnon to sig-cloud-provider-aws-maintainers

cartermckinnon:
Adding myself as a maintainer in order to create GitHub releases.


nckturner:
Oh looks like @elmiko isn't an approver. Will approve since its aws and probably should figure out who should be in that list. 

/approve

PR #4660, opened 2024-01-06
[1.30] Add Leads to Milestone Maintainers 

katcosgrove:
Adds RT lead shadows and subteam leads to milestone maintainers, and enhancements lead to the SIG Arch team.

Removes members from previous cycles.


katcosgrove:
cc: @kubernetes/sig-release-leads

PR #4709, opened 2024-01-23
Remove tnozicka from sig-apps teams

tnozicka:
Removing myself from sig-apps teams as I am no longer actively working there.


mrbobbytables (commented):
/lgtm
/approve

PR #4742, opened 2024-02-07
update sig-network-leads team

danwinship:
Antonio and I were added to the `sig-network-leads` _group_ but not the `sig-network-leads` _team_ I guess? Which apparently means we can't set `lead-opted-in` on KEPs? (https://github.com/kubernetes/enhancements/issues/784#issuecomment-1924751192)

/assign @thockin 
cc @shaneutt @MikeZappa87 @aojea

### DIFF
--- a/config/kubernetes/sig-network/teams.yaml
+++ b/config/kubernetes/sig-network/teams.yaml
@@ -77,6 +77,8 @@ teams:
   sig-network-leads:
     description: ""
     members:
+    - aojea
+    - danwinship
     - MikeZappa87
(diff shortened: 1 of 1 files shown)


aojea:
>     config_test.go:268: The team sig-network-leads in org Kubernetes has an unsorted list of members

lol

thockin:
Thanks!

/lgtm
/approve

aojea:
> ah, I was assuming the ASCII ordering was intentional...

I've found out today too,  I was assuming the same :)

Issue #4754, opened 2024-02-13
REQUEST: Migrate jpbetz/auger to etcd-io/auger


jmhbnz:
### New repo, staging repo, or migrate existing

Migrate existing

### Is it a staging repo?

no

### Requested name for new repository

auger

### Which Organization should it reside

etcd-io

### Who should have write access?

Refer existing OWNERS in jpbetz/auger

### Who should be listed as approvers in OWNERS?

Refer existing OWNERS in jpbetz/auger

### Who should be listed in SECURITY_CONTACTS?

Refer existing SECURITY_CONTACTS in jpbetz/auger

### What should the repo description be?

Directly access data objects stored in etcd by kubernetes.

### What SIG and subproject does this fall under?

sig-etcd

### Please provide references to appropriate approval for this new repo

https://github.com/jpbetz/auger/issues/42

jmhbnz:
cc @wenjiaswe, @siyuanfoundation, @ahrtr, @serathius, @jpbetz

serathius:
+1

ahrtr:
+1

Thanks @jmhbnz for driving this.

wenjiaswe:
+1 Thanks!

siyuanfoundation:
+1. Thanks James for leading this!

jpbetz:
+1

Thank you all for finding a home for auger.

mrbobbytables:
/assign
@jpbetz I'll invite you to our intermediate org to transfer the repo there before we can move it to etcd. 👍

mrbobbytables:
Repo has been migrated - I've opened a PR to add the subproject here:  https://github.com/kubernetes/community/pull/7718
@jmhbnz  opened this PR to add the teams: https://github.com/kubernetes/org/pull/4755
Once both are merged, this should be good to close 👍

wenjiaswe:
Thank you so much @mrbobbytables !

PR #4755, opened 2024-02-13
Added team for etcd-io/auger

jmhbnz:
Fixes https://github.com/kubernetes/org/issues/4754

/hold

Hold pending repo migration.

### DIFF
--- a/config/etcd-io/org.yaml
+++ b/config/etcd-io/org.yaml
@@ -107,7 +107,7 @@ teams:
     repos:
       raft: maintain
   maintainers-website:
-    description: ""
+    description: Granted write access to website
     members:
     - ahrtr
     - chalin
@@ -121,6 +121,15 @@ teams:
     privacy: closed
     repos:
       website: admin
+  maintainers-auger:
+    description: Granted write access to auger
+    members:
+    - jmhbnz
+    - siyuanfoundation
+    - wenjiaswe
+    privacy: closed
+    repos:
+      auger: maintain
   members:
     description: ""
     members:
@@ -154,3 +163,4 @@ teams:
           gofail: triage
           raft: triage
           website: triage
+          auger: triage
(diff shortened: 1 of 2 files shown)


mrbobbytables (changes requested):
To use the repo permissions features the repo needs to be added to the list here:

https://github.com/kubernetes/org/blob/25da6e3d0bbc154da53d998899dc8aa388d9eb33/config/restrictions.yaml#L376-L389

right now they're all pretty much just explicitly defined, but it should work for regex patterns.

jmhbnz:
Good catch @mrbobbytables - Should be sorted now 👍🏻

jmhbnz:
/unhold

Repo migration has now been completed, https://github.com/etcd-io/auger is live.

config/etcd-io/org.yaml
```diff
@@ -121,6 +121,15 @@ teams:
...
       website: admin
+  maintainers-auger:
+    description: Granted write access to auger
+    members:
+    - jmhbnz
+    - siyuanfoundation
+    - wenjiaswe
+    privacy: closed
+    repos:
+      auger: admin
```
MadhavJivrajani, line 132: Just making sure you meant `admin` here - the use of `admin` vs `maintain` is not consistent across etcd teams (might be worth changing that in the future!)

MadhavJivrajani (commented):
Just one comment, otherwise lgtm

config/etcd-io/org.yaml
MadhavJivrajani: If you did mean admin, I would also recommend reflecting that in the team description by s/write/admin

jmhbnz: Good catch, I'll confess I had just copied it from the above entry which said `admin`.  I think `maintain` should be fine as that seems to be used elsewhere.  We can probably change website to `maintain` but I would prefer to do that separately.

MadhavJivrajani (approved):
/lgtm
/approve

Thanks!

Issue #4726, opened 2024-01-28
Cleanup Kubernetes GitHub organizations - Jan 2024


palnabarun:
An org membership grants one access to an elevated set of permissions, and many members also grow to be owners of code. Being inactive for a long period of time impacts the project. We have a [policy](https://github.com/kubernetes/community/blob/master/community-membership.md#inactive-members) for defining who is an inactive member. In the past, we have periodically cleaned up inactive members from our several GitHub organizations.

Last year, we made some [amendments](https://github.com/kubernetes/community/pull/7380) to our guidelines for inactivity, notably reducing the period for measurement of inactivity to 12 months.

Previous Instances of cleanup: https://github.com/kubernetes/org/issues/2456, https://github.com/kubernetes/org/issues/2076

If you find yourself in the below list and contribute actively in an area not captured by Devstats, please comment here with details about your contributions. The deadline for exceptions is February 12, 2024. If you miss raising an exception, one can initiate a revert of their removal from the repo.

Following is the list of inactive members who are slated to be removed from the Kubernetes GitHub organizations:

```
AGMETEOR
AdamDang
Adirio
AlexeyPerevalov
Amotul-raheem
AndyXiangLi
Arvinderpal
BaluDontu
Birdrock
Bradamant3
Bubblemelon
CKchen0726
CaoDonghui123
ChandaniM123
ChiefAlexander
CindyXing
ClaudiaJKang
ConnorDoyle
Cweiping
Cynerva
DaiHao
Damans227
DanyC97
EleanorRigby
EmilyM1
EppO
G-Harmon
Garima-Negi
GinnyJI
Jerry-Ge
JieJhih
JohnTitor
KashifSaadat
KentaTada
KimDoubleB
KofClubs
Lion-Wei
Liujingfang1
Lp-Francois
LuckySB
MAKOSCAFEE
MHBauer
MJ-CJM
MarcusNoble
MaxRenaud
Miouge1
Monkeyanator
MonzElmasry
NayeonKeum
NickrenREN
NicoleYson
NilimaC04
Nivedita-coder
OddBloke
PatrickLang
Pensu
Phillels
Potapy4
Pothulapati
PriyankaH21
Promaethius
PurneswarPrasad
QiWang19
RCantw3ll
Rajakavitha1
Rajat-0
RenaudWasTaken
RobertKrawitz
SEJeff
ScorpioCPH
ScrapCodes
Shell32-Natsu
SlickNik
SyamSundarKirubakaran
[...]

PR #4758, opened 2024-02-13
Remove inactive members - 2024

palnabarun:
Remove inactive members identified in https://github.com/kubernetes/org/issues/4726 from the respective orgs and teams.

There are some more cleanups which would happen as a followup.


cmd/korg/korg.go
```diff
@@ -46,6 +46,15 @@ Add user to specified orgs:
 Note: Adding to teams is currently unsupported.
 	`
 
+	removeHelpText = `
+Remove users from GitHub orgs and/or teams
+
+Remove user to specified orgs:
+
+	korg remove <github username> --org kubernetes --org kubernetes-sigs
+	korg remove <github username> --org kubernetes --teams kubernetes/kubernetes-maintainers
```
Priyankasaggu11929, line 55: Both `korg remove/add ...` throws missing `--teams` flag. Is it a TODO or something to clean?

```
> korg remove Priyankasaggu11929 --org kubernetes --teams kubernetes/kubernetes-maintainers
Error: unknown flag: --teams
Usage:
  korg remove [flags]

```

cmd/korg/remove.go
```diff
@@ -0,0 +1,77 @@
...
+
+		config, err := readConfig(configPath)
+		if err != nil {
+			return fmt.Errorf("reading config: %s", err)
+		}
+
+		if !stringInSliceCaseAgnostic(config.Members, username) {
+			return fmt.Errorf("user %s doesn't exist in org %s", username, org)
+		}
+
```
Priyankasaggu11929, line 49: In case of admins, let's call out the username is an admin
(currently we get `Error: user <username> doesn't exist in org kubernetes`)

```suggestion
                if stringInSliceCaseAgnostic(config.Admins, username) {
                        return fmt.Errorf("user %s is an admin for org %s", username, org)
                }
                
		if !stringInSliceCaseAgnostic(config.Members, username) {
			return fmt.Errorf("user %s doesn't exist in org %s", username, org)
		}

```

cmd/korg/remove.go
```diff
@@ -0,0 +1,77 @@
...
+		configPath := filepath.Join(o.RepoRoot, relativeConfigPath)
+
+		config, err := readConfig(configPath)
+		if err != nil {
+			return fmt.Errorf("reading config: %s", err)
+		}
+
+		if !stringInSliceCaseAgnostic(config.Members, username) {
+			return fmt.Errorf("user %s doesn't exist in org %s", username, org)
+		}
```
MadhavJivrajani, line 48: Note for the future: once we bump to the latest go version, let's replace these helpers with the ones present in the sodlib slice package.

cmd/korg/remove.go
```diff
@@ -0,0 +1,77 @@
...
+
+		if o.Confirm {
+			fmt.Printf("saving config for %s org\n", org)
+			if err := saveConfig(configPath, config); err != nil {
+				return fmt.Errorf("saving config: %s", err)
+			}
+		}
+
+		configsModified = append(configsModified, relativeConfigPath)
```
MadhavJivrajani, line 65: Is the config always modified? IIUC, this sees a particular config as modified regardless of if a user matched for removal or not?

cmd/korg/remove.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	if o.Confirm {
```
MadhavJivrajani, line 68: We should probably also print out the modified configs here regardless of `--confirm` to help with dry running.

MadhavJivrajani (commented):
Thank you so much ❤️

MadhavJivrajani:
/hold 
for after the WIP is removed

cmd/korg/remove.go
palnabarun: Not always. Only when the user is present in the org. If the user doesn't match, we return early.

cmd/korg/korg.go
palnabarun: It was a remnant from what I originally planned the command to be.

cmd/korg/remove.go
palnabarun: This is a good suggestion. Added.

cmd/korg/remove.go
palnabarun: Printing to stdout will result in 180-1700 lines of debug info in the terminal, depending on which org is being modified.

palnabarun:
Thank you, @Priyankasaggu11929 and @MadhavJivrajani, for reviewing this quickly. ❤️ 

I have pushed the suggestions as a fixup commit. I will auto squash and rebase once all comments are resolved, as rebase takes a good amount of time with this many commits.

cmd/korg/remove.go
MadhavJivrajani: Makes sense - how about we print out just the config files that would be changed and not the entire diff?

palnabarun: done.

palnabarun:
Not sure why the needs rebase label is still there.

mrbobbytables:
Just so others know from a breadcrumb:
`needs-rebase` is no longer blocking, its a best effort signal (GitHub itself is the better source of truth) as it can be hit or miss from things like like quota exhaustion or missed webhooks

@michelle192837 followed up and verified that it was indeed throttled this morning.

All that said:
/lgtm
👍

PR #4833, opened 2024-03-18
Remove ehashman as SIG Instrumentation lead

ehashman:
I've removed myself from the alias/team for Instrumentation leads as well as repos I am not active in. POLA, etc.

/cc @kubernetes/sig-instrumentation-leads


dashpole:
/lgtm
/approve
for sig-instrumentation

cblecker (approved):

PR #4852, opened 2024-03-28
chore: do not require gobin to be in path

justinsb:
`go run` now works very well and avoids the path dependency.


justinsb:
Just a small nit I encountered while trying out the tooling here - otherwise it worked really well!

PR #4853, opened 2024-03-28
Add suggested affiliation check to membership template

mrbobbytables:
Adds 2 items to suggest that users ensure their affiliation is up to date. It is not required, but preferred.

/hold


soltysh (approved):
/lgtm

palnabarun (approved):
/lgtm

PR #4887, opened 2024-04-17
Use go install to install peribolos

cblecker:
Peribolos has moved within the prow repo to `sigs.k8s.io/prow/cmd/peribolos`. This fixes the job to use that new path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ SHELL := /usr/bin/env bash
 
 # available for override
 GITHUB_TOKEN_PATH ?=
-PROW_PATH ?= $(OUTPUT_DIR)/tmp/prow
 
 # intentionally hardcoded list to ensure it's high friction to remove someone
 ADMINS = cblecker MadhavJivrajani mrbobbytables nikhita palnabarun Priyankasaggu11929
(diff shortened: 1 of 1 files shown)


palnabarun (approved):
/lgtm
/approve

Makefile
```diff
@@ -81,10 +80,5 @@ $(MERGED_CONFIG): clean $(MERGE_CMD) $(CONFIG_FILES)
...
 
-$(PROW_PATH):
-	mkdir -p $(PROW_PATH)
-	git clone --depth=1 https://github.com/kubernetes-sigs/prow $(PROW_PATH)
-
-$(PERIBOLOS_CMD): $(PROW_PATH)
-	cd $(PROW_PATH) && \
-		go build -o $(PERIBOLOS_CMD) ./prow/cmd/peribolos
+$(PERIBOLOS_CMD):
+	GOBIN=$(OUTPUT_BIN_DIR) go install sigs.k8s.io/prow/cmd/peribolos@latest
```
BenTheElder, line 84: note: this actually means the most recent tag, if you want to continue installing the _latest_ code (equivalent to cloning without specifying a version) you need to do `@$branch`

https://go.dev/ref/mod#version-queries
https://go.dev/ref/mod#pseudo-versions

prow isn't tagged yet, but when/if it starts tagging this will change behavior following the PR here

PR #4888, opened 2024-04-17
Use main branch instead of latest

cblecker:
As @BenTheElder pointed out https://github.com/kubernetes/org/pull/4887#discussion_r1569209311, it's better for us to use the pseudo version for the main branch, as opposed to the "latest" tag, in the case that the prow repo starts using versioned tags in the future.


BenTheElder (commented):
/lgtm

PR #4941, opened 2024-05-10
docs: add salaxander as SIG docs tech lead

salaxander:
Adding salaxander as a SIG docs tech lead

CC @natalisucks @divya-mohan0209 @reylejano


natalisucks:
/lgtm
so excited to have you on board, @salaxander! 🫶

PR #4952, opened 2024-05-16
Add Okabe-Junya to kubernetes-sigs

Okabe-Junya:
## Feature Description

- Add @Okabe-Junya to kubernetes-sigs

## Note

I am already part of the Kubernetes org, so I add myself directly

> However, if you are already part of the Kubernetes organization, you do not need to do this and can add yourself directly to the appropriate files.


Okabe-Junya:
Thank you for your quick response!!

PR #5115, opened 2024-08-27
Remove failed org invites

cblecker:
This PR removes users who haven't accepted their org invites


config/etcd-io/org.yaml
```diff
@@ -26,17 +26,14 @@ members:
 - fanminshi
 - fuweid
 - gryf
-- hakman
```
jmhbnz, line 29: Just to clarify - are these invites now expired?  I.e. if the member did need access would a new pr need to be raised?

For `etcd-io`:
-  I agree that `mitake` should be removed as they are now emeritus maintainer. 
- `hakman` and `justinsb` are part of the new etcd operator working group as leads and will need access sorted now that our new `etcd-io/etcd-operator` repo has been created.

cblecker: They haven't accepted the invites out to them for multiple weeks.. this is a cleanup.

If they actually need this access, yes we can PR them back in and a new invite will go out. But if they don't accept it, they get cleaned back out. (folks need to accept an invite for each new org they join.. once they are in, team access can be added or removed without further user interaction)

cblecker: Oh and one last thing: GitHub gives you 7 days from when the invite is sent to accept it before it expires and is listed as a failed invitation.

jmhbnz: Thanks for confirmation.  

cc @hakman and @justinsb please raise a new `kubernetes/org` pr in future to add yourself to `etcd-io` org when needed for etcd operator wg lead duties.

hakman: @cblecker @jmhbnz I think both me and @justinsb accepted the invites. I could even see the org in my profile. Not sure what happened there.

hakman: New PR created: https://github.com/kubernetes/org/pull/5116

PR #5125, opened 2024-09-02
Add 1.32 release team permissions

fsmunoz:
xref: https://github.com/kubernetes/sig-release/issues/2603

cc @katcosgrove @gracenng


gracenng:
I believe @mickeyboxell will be RT Branch Manager Shadow if you want to add that as well

fsmunoz:
@gracenng indeed, fixed in latest commit, thanks!

PR #5150, opened 2024-09-12
Update teams.yaml with release notes shadows 1.32

satyampsoni:
cc @fsmunoz @katcosgrove @gracenng


gracenng:
/lgtm
/approve
Can't merge this until new folks are member - feel free to tag Kat and I on those issues

fsmunoz:
@satyampsoni there's a dependency on their org status, but I think you also have a sorting problem that you can take care of while waiting.

Issue #5097, opened 2024-08-08
REQUEST: Migrate elmiko/karpenter-provider-cluster-api


elmiko:
### New repo, staging repo, or migrate existing

migrate existing elmiko/karpenter-provider-cluster-api

### Is it a staging repo?

no

### Requested name for new repository

karpenter-provider-cluster-api

### Which Organization should it reside

kubernetes-sigs

### Who should have admin access?

elmiko, fabriziopandini, sbueringer, vincepri

### Who should have write access?

elmiko, chrischdi, enxebre, fabriziopandini, killianmuldoon, sbueringer, vincepri

### Who should be listed as approvers in OWNERS?

elmiko, chrischdi, enxebre, fabriziopandini, killianmuldoon, sbueringer, vincepri

### Who should be listed in SECURITY_CONTACTS?

elmiko, fabriziopandini, vincepri

### What should the repo description be?

Cluster API Karpenter provider. Karpenter is a Kubernetes Node Autoscaler built for flexibility, performance, and simplicity. 

### What SIG and subproject does this fall under?

sig cluster-lifecycle, sub-project cluster-api

### Please provide references to appropriate approval for this new repo

The naming and migration of the repo has been discussed with the [SIG Autoscaling community on 2024-01-15][meet1], and the [Karpenter community on 2024-01-18][meet2], and most recently in the [Cluster API community on 2024-08-07][meet3].

[meet1]: https://docs.google.com/document/d/1RvhQAEIrVLHbyNnuaT99-6u9ZUMp7BfkPupT2LAZK7w/edit
[meet2]: https://docs.google.com/document/d/18BT0AIMugpNpiSPJNlcAL2rv69yAE6Z06gUVj7v_clg/edit
[meet3]: https://docs.google.com/document/d/1GgFbaYs-H6J5HSQ6a7n4aKpk0nDLE2hgG2NSOM9YIRw/edit#bookmark=id.hoco8dz1s099

### Additional context for request

This topic will also be discussed after a consensus period on 20 August at the next SIG Cluster Lifecycle meeting.

elmiko:
happy for comments on this PR, but we don't want to merge this until after the SIG cluster lifecycle meeting on 20 August 2024.

/hold until 2024-08-20

vincepri:
+!

chrischdi:
+1

neolit123:
+1

neolit123:
github team, we discussed that we need consensus from the sig leads before the migration.

[...]

PR #5183, opened 2024-09-25
sig-cluster-lifecycle: add karpenter-provider-cluster-api github teams

Priyankasaggu11929:
ref: https://github.com/kubernetes/org/issues/5097

/assign @kubernetes/sig-cluster-lifecycle-leads

cc: @kubernetes/owners


neolit123 (commented):
it matches https://github.com/kubernetes/org/issues/5097 thanks @Priyankasaggu11929 !

/lgtm
/approve

PR #5199, opened 2024-10-02
Update Steering members following 2024 election cycle

justaugustus:
cc: @kubernetes/steering-committee 
ref: https://github.com/kubernetes/steering/issues/286

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -169,12 +169,12 @@ aliases:
     - ritazh
     - tabbysable
   committee-steering:
+    - aojea
     - BenTheElder
     - justaugustus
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -170,11 +170,11 @@ aliases:
     - tabbysable
   committee-steering:
     - BenTheElder
+    - aojea
```
BenTheElder, line 173: non-blocking suggestion: use the same sort as in org.yaml?

BenTheElder (commented):
thanks, LGTM with one comment

saschagrunert (approved):
Thank you!

Issue #5280, opened 2024-12-01
REQUEST: Create kubernetes-sigs/ingate


strongjz:
### New repo, staging repo, or migrate existing

ingate

### Is it a staging repo?

yes

### Requested name for new repository

ingate

### Which Organization should it reside

kubernetes-sigs

### Who should have admin access?

strongjz rikatz Gacko

### Who should have write access?

strongjz rikatz Gacko tao12345666333

### Who should be listed as approvers in OWNERS?

cpanato  Gacko  puerco   rikatz  strongjz   tao12345666333

### Who should be listed in SECURITY_CONTACTS?

Gacko rikatz strongjz

### What should the repo description be?

InGate - an Ingress & Gateway API Controller

### What SIG and subproject does this fall under?

sig-network

### Please provide references to appropriate approval for this new repo

- Sig-networking mailing list https://groups.google.com/g/kubernetes-sig-network/c/klXafeV2lPA/m/ZgLIW0n0BgAJ
- Gateway API Community meeting https://youtu.be/cW9YQQ3Cuc8?feature=shared&t=2294
- Ingress-nginx maintainer talk  https://www.youtube.com/watch?v=KLwsV6_DntA

shaneutt:
:+1:

cpanato:
+1

mrbobbytables:
Repo has been created: https://github.com/kubernetes-sigs/ingate
once https://github.com/kubernetes/org/pull/5281 and #8200 merge you should be good to go and this can be closed :+1:

aojea:
I think we can close this now, I see both related PR are merged now

Thanks all

/close

PR #5281, opened 2024-12-02
Add teams for ingate repo

mrbobbytables:
ref: #5280

/assign @strongjz @rikatz @Gacko


strongjz (approved):
/lgtm

Gacko:
Hey!

Thank you for taking care of this. Just one question: My handle is Gacko with an upper G. Do you think this matters somewhere as you added it with a lower g?

PR #5307, opened 2024-12-16
Add serngawy to kubernetes-sig org

serngawy:
The PR fixes  https://github.com/kubernetes/org/issues/5306


Priyankasaggu11929:
/close

Hello @serngawy, thanks for the PR.

But the [membership request](https://github.com/kubernetes/org/issues/5306) needs to be first reviewed by the New Membership Coordinator and then will be processed in next batch. Thanks!

Closing per ^

Issue #5302, opened 2024-12-10
REQUEST: Migrate llm-instance-gateway to gateway-api-inference-extension


robscott:
### New repo, staging repo, or migrate existing

Migrate existing

### Is it a staging repo?

no

### Requested name for new repository

gateway-api-inference-extension

### Which Organization should it reside

kubernetes-sigs

### Who should have admin access?

No change

### Who should have write access?

No change

### Who should be listed as approvers in OWNERS?

No change

### Who should be listed in SECURITY_CONTACTS?

No change

### What should the repo description be?

No change

### What SIG and subproject does this fall under?

SIG Network (see https://github.com/kubernetes/community/pull/8211)

### Please provide references to appropriate approval for this new repo

Lots of different discussions around this, but https://github.com/kubernetes/community/pull/8211 has the most recent one.

robscott:
SIG Net Leads:
/cc @aojea @danwinship @MikeZappa87 @shaneutt @thockin

WG Serving Leads:
/cc @ArangoGutierrez @Jeffwan @SergeyKanzhelev @terrytangyuan

aojea:
I didn't know it was possible to rename a repo

/lgtm

BenTheElder:
> I didn't know it was possible to rename a repo

Yes, and Github will actually serve redirects as long as the old name isn't reused.

https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository

terrytangyuan:
@BenTheElder Who should be the one actually doing the renaming for that repo?

Jeffwan:
it has been discussed.

/lgtm

danehans:
>Yes, and Github will actually serve redirects as long as the old name isn't reused.

To ensure this does not happen, what do you think about the following steps:

1. Create a new gateway-api-inference-extension repo.
2. Move applicable llm-instance-gateway repo content to gateway-api-inference-extension, except for the README stating the repo is retired with a link to the gateway-api-inference-extension repo.
3. After some time, remove the llm-instance-gateway repo.

aojea:
we own the process so for someone to reuse the name has to ask for approval first

BenTheElder:
> @BenTheElder Who should be the one actually doing the renaming for that repo?

The owners of this repo (`@kubernetes/owners`) https://github.com/kubernetes/community/blob/5ed4178ced364351ce4f4e485e7f332c7a481937/github-management/README.md#github-administration-team

[...]

PR #5309, opened 2024-12-17
k-sigs: rename and move inference project from sig-apps to sig-net

MadhavJivrajani:
Ref https://github.com/kubernetes/org/issues/5302 and https://github.com/kubernetes/community/pull/8211


BenTheElder (commented):
/lgtm

PR #5328, opened 2025-01-07
Add sig-autoscaling-milestone-maintainers

adrianmoisey:
The VPA project would like to start using milestones to track PRs and issues for review. In order to do so we need access to manage milestones.

cc'ing @gjtempleton, who is the chair of sig-autoscaling

Note: I'm unsure if this is the correct process, please point me in the right direction if it is not


gjtempleton:
Thanks!
/lgtm
/approve

PR #5349, opened 2025-01-18
fix: update sig-scheduling-misc team

sanposhiho:
This PR updates sig-scheduling-misc team members based on the current reviewers/approvers:
https://github.com/kubernetes/kubernetes/blob/master/OWNERS_ALIASES

I added @dom4ha as well since I think will join the list soon ;)


AxeZhan:
/lgtm 

Great!!!

Huang-Wei (commented):
/lgtm
/approve

PR #5373, opened 2025-01-25
add sttts to sig-api-machinery-leads

BenTheElder:
I noticed the discrepancy when going to ping on https://github.com/kubernetes/test-infra/pull/34210

cc @sttts @kubernetes/sig-api-machinery-leads


jpbetz:
/lgtm
Good catch

palnabarun:
Approving since it is lgtm'ed by a lead and the change is syncing teams to actual source of truth.

/approve

PR #5387, opened 2025-01-29
Update approvers and reviewers dropping inactive members and adding shadows

soltysh:
/assign @liggitt


liggitt:
/lgtm
from apireview side

thanks

PR #5395, opened 2025-02-10
Add hime as admin to gcp-compute-persistent-disk-csi-driver

hime:
Add admin permissions to @hime to [gcp-compute-persistent-disk-csi-driver](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver) admin.


msau42:
/lgtm
/approve

Jaime is part of the release team for pdcsi driver.

It is strange that the driver is under sig-storage, it should be under sig-cloud-provider with all the other csi drivers, but we can move that in a followup pr.

PR #5438, opened 2025-03-01
Give maintainers-website permissions for etcd-io/protodoc

jmhbnz:
So that we can maintain the proto docs that get generated and hosted on https://etcd.io

cc @ahrtr, @serathius, @ivanvc


ahrtr:
/lgtm
/approve

Thanks

PR #5445, opened 2025-03-05
etcd: Update release-etcd team for release window

ivanvc:
Add myself to the release team in preparation for today's releases.

/cc @jmhbnz @ahrtr

### DIFF
--- a/config/etcd-io/sig-etcd/teams.yaml
+++ b/config/etcd-io/sig-etcd/teams.yaml
@@ -157,7 +157,7 @@ teams:
     description: Granted permission to release etcd-io/etcd
     # The member list is intentionally empty. It should only include the release
     # lead during release windows.
-    members: []
+    members: [ivanvc]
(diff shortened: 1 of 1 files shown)


ahrtr:
@ivanvc just keep you in the list. No need to remove it after the release.

ivanvc:
> @ivanvc just keep you in the list. No need to remove it after the release.

We prefer not to leave it as-is because it would be a backdoor for the maintainer's permission. I'll open a pull request to restore it. I'll also submit this with more time to avoid rushing it.

ahrtr:
> We prefer not to leave it as-is because it would be a backdoor for the maintainer's permission

I don't think it's backdoor. You are the release lead, so should have the write permission to the repo.

ahrtr:
> I don't think it's backdoor. You are the release lead, so should have the write permission to the repo.

I think we can add both you and @jmhbnz to the list.

PR #5489, opened 2025-03-27
id: added za as approvers and reviewers for id content

za:
https://github.com/kubernetes/org/issues/5488


Priyankasaggu11929:
cc: @kubernetes/sig-docs-leads 


/ok-to-test

girikuncoro:
Looking good! we need @za to keep sig-docs-id contributions keep coming

/lgtm
/approve

PR #5500, opened 2025-04-01
Add member 2025-04-01

truongnh1992:
Fixes https://github.com/kubernetes/org/issues/5480

/area github-membership

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -494,6 +494,7 @@ members:
 - kflynn
 - kfox1111
 - kfswain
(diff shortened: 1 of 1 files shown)


mrbobbytables:
@truongnh1992  thanks for opening the PR, but we have a specific format we follow for the commit message when adding/removing people from the org^^;; 

I've added them in the org update PR here: #5522
Once that merges, they should get their invite :+1: 

With that I'm going to go ahead and close this one out for now.
Thanks!

khanhtc1202:
@mrbobbytables Thanks for the update 😄

PR #5520, opened 2025-04-09
Add rthallisey as a member

rthallisey:
fixes: #5518


mrbobbytables:
@rthallisey thanks for opening the PR, but we have a specific format we follow for the commit message when adding/removing people from the org^^;; 

I've added you to org membership additions in PR #5522, so you should be taken care of there :+1: 

With that I'm going to go ahead and close this one out for now.

PR #5360, opened 2025-01-21
Elevate etcd team permissions for release window

jmhbnz:
Tomorrow we are releasing etcd `v3.5.18` .

Part of: 
- https://github.com/etcd-io/etcd/issues/19119

Elevated permissions required as per https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/release.md#release-steps

Permissions will be returned to normal after release completed by a subsequent pr.

cc @serathius, @ahrtr, @ivanvc, @ghouscht

PR #5523, opened 2025-04-09
Create a separate group with etcd admin access

serathius:
I'm working on Antithesis integration as proposed in https://github.com/etcd-io/etcd/issues/19299.

We would like to setup a github workflow that calls a third party service with credentials. This should be doable by setting up a secret in github actions as described in https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions. At the moment no etcd maintainer has permissions to do that. 

I get permission I copied the the etcd-operator-admins group and I would like to give `admin` access to SIG etcd leads. 
Can be temporary just for the secret setup.

/cc @ahrtr @jmhbnz @ivanvc @wenjiaswe

### DIFF
--- a/config/etcd-io/sig-etcd/teams.yaml
+++ b/config/etcd-io/sig-etcd/teams.yaml
@@ -1,4 +1,16 @@
 teams:
+  etcd-admins:
+    description: Admin access to etcd repo
+    members:
+    - ahrtr
+    - fuweid
+    - jmhbnz
+    - serathius
+    - spzala
+    - wenjiaswe
+    privacy: closed
+    repos:
+      etcd: admin
   etcd-operator-admins:
     description: Admin access to etcd-operator repo
     members:


ahrtr:
/lgtm
/approve



> Can be temporary just for the secret setup.

Do you mean the team `etcd-admins` will be removed once it isn't needed anymore (e.g. after the secret setup is done)?

ahrtr:
Oh, I should hold the PR until other maintainers review :(

jmhbnz:
Just a note to confirm it is intentional that from a day to day perspective we operate without `admin` privilege on repositories.  

When situations like this pop up where we require `admin` we historically have just temporarily bumped the permissions for `maintainers-etcd` group to move it from `maintain` ---> `admin` and then revert the commit once neccessary changes are complete.  Example is: https://github.com/kubernetes/org/pull/5360

No issues with this pr - let's just revert the new group once any necessary changes to enable Antithesis integration have wrapped up and return to our least privilege operating principle.

PR #5528, opened 2025-04-10
Add missing maintainers + headlamp-reviewers team to the sig-ui teams

joaquimrocha:
I know normally reviewers do not have write permissions in the repo, but in this case the headlamp core team is used to push branches directly, so it's more convenient than having forks outside.

### DIFF
--- a/config/kubernetes-sigs/sig-ui/teams.yaml
+++ b/config/kubernetes-sigs/sig-ui/teams.yaml
@@ -10,7 +10,8 @@ teams:
     description: Write access to the headlamp repo
     members:
     - joaquimrocha
+    - sniok
+    - illume
     privacy: closed
     repos:
-      headlamp: write
-  
+      headlamp: maintain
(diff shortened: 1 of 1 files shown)


illume:
The failing test is because one of our members isn't in the k8s org yet.
>     config_test.go:268: The following members of team headlamp-reviewers are not Kubernetes SIGs org members: yolossn

@mrbobbytables Can you please add yolossn to Kubernetes SIGs org?
- https://github.com/kubernetes/org/issues/5511

config/kubernetes-sigs/sig-ui/teams.yaml
```diff
@@ -10,7 +10,19 @@ teams:
     description: Write access to the headlamp repo
     members:
     - joaquimrocha
+    - sniok
+    - illume
```
Priyankasaggu11929, line 14: Needs sorting the members list in alphabetical order here:

```
    - illume
    - joaquimrocha
    - sniok
```

Priyankasaggu11929 (changes requested):

config/kubernetes-sigs/sig-ui/teams.yaml
joaquimrocha: Completely forgot about this. Sorry. Will get it done now.

PR #5574, opened 2025-05-05
Add sunnylovestiramisu as admin for volume populators

sunnylovestiramisu:
Need permission to create release/tags

/assign @xing-yang

### DIFF
--- a/config/kubernetes-csi/org.yaml
+++ b/config/kubernetes-csi/org.yaml
@@ -564,6 +564,7 @@ teams:
     members:
     - bswartz
     - saad-ali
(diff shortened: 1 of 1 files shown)


config/kubernetes-csi/OWNERS
```diff
@@ -13,6 +13,7 @@ approvers:
   - nikhita
   - palnabarun
   - Priyankasaggu11929
+  - sunnylovestiramisu
```
xing-yang, line 16: You can't add yourself here.  These are leads from sig-contribex

PR #5578, opened 2025-05-06
Include Prajyot-Parab as an admin for cluster-api-provider-ibmcloud

mkumatag:
@Prajyot-Parab is one of the active member and approver for the project, hence adding him as an admin to the repository for an additional maintenance required for the project.


spzala (approved):
+1 Thank you @mkumatag @Prajyot-Parab 
/lgtm

PR #5579, opened 2025-05-06
Update network-policy-api admins

danwinship:
@astoycos has been out of the picture for a while and @tssurya and @npinaeva have been running the WG, so add them to the admin group so they can do releases and such. (Also add @aojea since he's an approver there.)


aojea:
/lgtm
/approve

Thanks

PR #5580, opened 2025-05-07
Add joshjms to etcd member

joshjms:
ref: https://github.com/kubernetes/org/pull/5567

/cc @ivanvc


ivanvc (approved):
/lgtm

Thanks, @joshjms.

/cc @sig-etcd-leads

PR #5671, opened 2025-06-23
add nirrozenbaum as owner of gateway api inference extension

nirrozenbaum:
(instead of jeffwan)


shaneutt (commented):
cc @ahg-g @danehans @kfswain

PR #5676, opened 2025-06-24
Add Wendy Ha to Etcd Org.

jberkus:
Wendy Ha is currently a Docs mentee for Etcd, and is already a Kubernetes member.  We need her to be an etcd member as well, so that she can interact with the bots.

attn: @nate-double-u @ivanvc @siyuanfoundation


jberkus:
Gah, apparently my editor did yaml reformatting I wasn't aware of.  Closing this until I can figure out how to disable that.

PR #5690, opened 2025-07-03
sig-cl: remove the general, outdated sig-cluster-lifecycle team

neolit123:
Historically this was a team with arbitrary membership. Majority of the users in this team are no longer active. Instead we should only maintain the [-leads team](https://github.com/kubernetes/org/pull/5690/files#diff-ce49e15ad0c55db9a5f2db9582f25d5fbc2fa9682d58cd12925cc527b6e76512R83) and we can be pinged from it.

/cc @fabriziopandini @justinsb @vincepri

### DIFF
--- a/config/kubernetes/sig-cluster-lifecycle/teams.yaml
+++ b/config/kubernetes/sig-cluster-lifecycle/teams.yaml
@@ -64,21 +64,6 @@ teams:
     privacy: closed
     repos:
       minikube: write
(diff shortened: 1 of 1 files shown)


justinsb:
Agree, I'm not even sure what this is for and think -leads would be the right choice in all cases I can think of.

The title is a little inflammatory :-)

neolit123:
looking for lgtm

PR #5692, opened 2025-07-07
add taylorwaggoner to kubernetes

mrbobbytables:
Adding Taylor to the kubernetes org so that she may be added as a billing manager in the kubernetes enterprise account.

/hold

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1154,6 +1154,7 @@ members:
 - tariq1890
 - tasdikrahman
 - tashimi
+- taylorwaggoner
 - tedyu
 - tengqm
 - tenzen-y


Priyankasaggu11929:
@mrbobbytables - should we add the username to the https://github.com/kubernetes/org/blob/main/docs/sample-exceptions.csv as well - to avoid removal during org membership audits?

mrbobbytables:
> @mrbobbytables - should we add the username to the https://github.com/kubernetes/org/blob/main/docs/sample-exceptions.csv as well - to avoid removal during org membership audits?

definitely - she wouldn't have any direct activity that shows up :+1:

MadhavJivrajani (approved):

PR #5758, opened 2025-08-05
sig-cl: sync system-validators admins/maintainers

neolit123:
sync with the k/kubeadm OWNERS file, so that anyone can cut system-validators releases.
also remove Fabrizio from k/kubeadm admins, but can be re-added if he wishes.

/cc @pacoxu @SataQiu


neolit123:
cc  @carlory @HirazawaUi

PR #5840, opened 2025-09-16
Add DamianSawicki and kl52752 as ingress-gce maintainers

swetharepakula:
/assign @bowei 
/assign @mmamczur

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1324,6 +1324,8 @@ teams:
     members:
     - aojea
     - bowei
+    - DamianSawicki
+    - kl52752 
     - mmamczur
     - MrHohn
     - swetharepakula


bowei:
/lgtm

but we should really move these owner entries into a different more focused owner directory and avoid having this in the main file

mmamczur:
we moved cloud-provider-gcp setup to https://github.com/kubernetes/org/tree/main/config/kubernetes/provider-gcp
so maybe something similar or even use that directory?

/lgtm

cblecker:
/hold
Oversight for this repo is under sig-network, so please move it under that directory if you are planning on moving it.

swetharepakula:
I updated the PR by moving the new ingress-gce directory into the sig-network directory

swetharepakula:
@cblecker, any update? Is that change match with what you were suggesting? Or should I keep it in the main config file?

swetharepakula:
@cblecker , please take a look. I have 2 maintainers of ingress-gce who have also lgtm'd the new change

aojea:
@swetharepakula @cblecker something is wrong now, I lost all the permissions on the ingress-gce repo.
Is there some dependency on the file or path that we ignored?

mmamczur:
> @swetharepakula @cblecker something is wrong now, I lost all the permissions on the ingress-gce repo. Is there some dependency on the file or path that we ignored?

We're looking at it

I tracked down how this machinery works, apparently there is a tool that syncs this file with github, called peribolos. It is run by prow and if you look at a recent run https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-org-peribolos/1996398641333407744 and search logs for ingress-gce:
time="2025-12-04T01:58:42Z" level=info msg="Skipping config/kubernetes/sig-network/ingress-gce and its children"

https://github.com/uwu-tools/peribolos/blob/676537a56f7866e1b1e6b2b9be7e0eee98d819ac/options/merge/merge.go#L120
this code will ignore the confid if the directory is nested too much. If you check other configs it is always kubernetes(-sig)/something/teams.yaml but never kubernetes(-sig)/something/second-something/teams.yaml props to @TortillaZHawaii for finding the relevant code

TortillaZHawaii:
I will remove the second-something part, @aojea which one should it be:
a) `config/kubernetes/ingress-gce/`
b) `config/kubernetes/sig-network/`

aojea:
everything should go always under a sig, there can not be independent projects that do not belong to a sig

> config/kubernetes/sig-network/

PR #5850, opened 2025-09-19
Add members sept-19-2025

Priyankasaggu11929:
fix: https://github.com/kubernetes/org/issues/5847
fix: https://github.com/kubernetes/org/issues/5843

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -795,6 +795,7 @@ members:
 - Sunnatillo
 - sunnylovestiramisu
 - sunpa93
+- sunya-ch
 - SwathiR03
 - swatisehgal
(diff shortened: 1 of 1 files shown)


config/kubernetes-sigs/org.yaml
```diff
@@ -795,6 +795,7 @@ members:
 - Sunnatillo
 - sunnylovestiramisu
 - sunpa93
+- sunya-ch
```
ahg-g, line 798: Can we pls add kaushikmitr to kubernetes-sigs as well, Kaushik is active there.

ahg-g: or is it the case that k8s members are also kubernetes-sigs members by default?

Issue #5848, opened 2025-09-18
REQUEST: Create `wg-ai-gateway` repo


nirrozenbaum:
### New repo, staging repo, or migrate existing

New repository

### Is it a staging repo?

no

### Requested name for new repository

wg-ai-gateway

### Which Organization should it reside

kubernetes-sigs

### Who should have admin access?

nirrozenbaum, keithmattix, kflynn, kfswain, shaneutt, xunzhuo

### Who should have write access?

nirrozenbaum, keithmattix, kflynn, kfswain, shaneutt, xunzhuo

### Who should be listed as approvers in OWNERS?

nirrozenbaum, keithmattix, kflynn, kfswain, shaneutt, xunzhuo

### Who should be listed in SECURITY_CONTACTS?

nirrozenbaum, keithmattix, kflynn, kfswain, shaneutt, xunzhuo

### What should the repo description be?

Proposals and discussions for the [AI Gateway Working Group](https://github.com/kubernetes/community/tree/master/wg-ai-gateway).

### What SIG and subproject does this fall under?

SIG Network, wg-ai-gateway

### Please provide references to appropriate approval for this new repo

Proposals and discussions and for the AI Gateway working group, which is sponsored by SIG Network and SIG Multi Cluster.

shaneutt:
+1 [other WGs](https://github.com/search?q=org%3Akubernetes-sigs+%2Fwg-%2F&type=repositories) have found it helpful to have a repository for their proposals, this would be helpful for us as well.

keithmattix:
+1

Xunzhuo:
+1

Priyankasaggu11929:
/assign

New repo is created – https://github.com/kubernetes-sigs/wg-ai-gateway

GitHub teams for the new repo – https://github.com/kubernetes/org/pull/5873

k/community documentation - https://github.com/kubernetes/community/pull/8632

Once above PRs are merged, the repo creation request is complete. Thanks!

Priyankasaggu11929:
closing as completed. Thanks folks!

/close

PR #5873, opened 2025-09-26
sig-network: add gh teams for wg-ai-gateway repo

Priyankasaggu11929:
ref: https://github.com/kubernetes/org/issues/5848

/assign @kubernetes/sig-network-leads 

cc: @kubernetes/owners


cblecker (approved):

Priyankasaggu11929:
I think we're only missing a `lgtm` label here. 

Could one of the tagged/assigned folks please add that too - thanks!

Issue #5881, opened 2025-09-29
Update elieser1101 to elieserr membership


elieserr:
### Organization or Repo

kubernetes, kubernetes-sigs

### Describe the issue

Recently changed my github handle so would like to update the user part of the orgs. The old handle was `elieser1101` and the new one is `elieserr`

The original org membership request is https://github.com/kubernetes/org/issues/5347

My understanding is that don't need to provide org request again, but would be happy to provide evidence of contributions + sponsors if needed.

/area github-membership

jasonbraganza:
Crossverified with the linked PR: #5347

PR #5885 should send you an invite, once it gets merged

Welcome back @elieserr :)

PR #5884, opened 2025-09-29
Update elieser1101 to elieserr membership

elieserr:
Recently changed my github handle so would like to update the user part of the orgs. The old handle was `elieser1101` and the new one is `elieserr`

The original org membership request is https://github.com/kubernetes/org/issues/5347

My understanding is that don't need to provide org request again, but would be happy to provide evidence of contributions + sponsors if needed.

/area github-membership

fixes https://github.com/kubernetes/org/issues/5881


jasonbraganza:
/close

@elieserr, have raised #5885 to close #5881

PR #6058, opened 2026-01-14
Add v1.36 Release Team members

rytswd:
This only updates the Release Lead, Release Lead Shadows, and Subteam Leads entries, as well as subteam members.

As I saw many names from v1.34 or before in the `milestone-maintainers`, I kept the list as is, and will probably look to clean up in a separate PR.


justaugustus (changes requested):
@rytswd — Minor comment nit to fix.

PR #6067, opened 2026-01-19
Move mrbobbytables out of blog owners

lmktfy:
Remove @mrbobbytables from our source of truth for @kubernetes/sig-docs-blog-owners. This change is OK because:
- we have other folks happy to help here
- Bob / @mrbobbytables has plenty of other things going on

Also see https://github.com/kubernetes/website/pull/54074

@mrbobbytables if you would LGTM this one, I'd appreciate it, just so we have on record that you're happy to step away.


mrbobbytables:
/lgtm

thanks!

PR #6113, opened 2026-02-02
Add ronaldngounou to kubernetes-sigs member

ronaldngounou:
I am already member of kubernetes org.

Add ronaldngounou to kubernetes-sigs member

```
We are currently working on automation that would transfer membership in the Kubernetes organization to any related orgs automatically, but such is not the case currently.
```

Issue https://github.com/kubernetes/org/issues/6109#issuecomment-3815619143

cc @jasonbraganza


jasonbraganza:
/close

@ronaldngounou, closing this because I raised #6116, looking at #6109

PR #6123, opened 2026-02-09
Add Patrick as tech lead for sig-instrumentation

richabanker:
Issue: https://github.com/kubernetes/community/issues/8818


dims:
yikes, i don't have karma here

/hold cancel

PR #6137, opened 2026-02-13
Add JoelSpeed and everettraven as api-reviewers

JoelSpeed:
Myself and @everettraven have recently been helping out with the API review team. Adding ourselves here so that we can get write access to the API review project board

CC @liggitt

### DIFF
--- a/config/kubernetes/org.yaml
+++ b/config/kubernetes/org.yaml
@@ -1170,6 +1170,8 @@ teams:
     members:
     - deads2k
     - enj
+    - everettraven
+    - JoelSpeed
     - jpbetz
     - liggitt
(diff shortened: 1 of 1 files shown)


JoelSpeed:
Huh, @everettraven we should get you set up as a member of the K8s org

liggitt:
+1 to both, can follow up to add Bryce or wait until his org add goes through, whichever you prefer

palnabarun:
@JoelSpeed -- since Bryce is already a k-sigs member, feel free to add a commit before your teams update to add Bryce to kubernetes org.

Commit message format: "add everettraven to kubernetes"

PR #6139, opened 2026-02-13
Offboarding cji

cji:
see: https://github.com/kubernetes/committee-security-response/issues/218

I am stepping down from the SRC. Thanks all! 👋🏻


cji:
(based on nikhita's status message, maybe I should re-assign?)

/assign @jasonbraganza

PR #6153, opened 2026-02-19
match current agentic-net leads

LiorLieberman:
/cc @david-martin @guicassolato


palnabarun:
ping @kubernetes/sig-network-leads for approval

PR #6268, opened 2026-04-03
Migrate GCP permissions

hajiler:


msau42:
/lgtm cancel

Sorry I got it wrong, I think it's https://github.com/kubernetes/org/blob/4ecb20cca1ca3716e2d01d68d98f90d6bef45ae2/config/kubernetes/provider-gcp/teams.yaml#L4 we want to move it to

msau42:
But I think the teams file has to be under the right org. So maybe we need to create a provider-gcp folder under kubernetes-sigs.

PR #6292, opened 2026-04-15
Update structured-merge-diff admins / maintainers

liggitt:
Sync with current https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/OWNERS


liggitt:
cc @jpbetz

PR #6398, opened 2026-05-24
korg: add userinfo command for org and OWNERS lookup

palnabarun:
Adds `korg userinfo <user>...` to answer the recurring question this repo
exists for: which Kubernetes orgs is a user in, with what role, and where
do they appear in OWNERS files.

Sources:
- GitHub user API for company / profile
- config/<org>/org.yaml (Members / Admins) for k8s org membership
- cs.k8s.io (hound) for OWNERS / OWNERS_ALIASES references

Behavior:
- Honors GITHUB_TOKEN / GH_TOKEN to avoid the 60/hr unauth limit. GitHub
  and hound use separate http.Client instances, so the token is never
  sent to cs.k8s.io
- Per-request context with a 30s timeout; SIGINT cancels in flight
- Multi-user batches run concurrently (errgroup, bounded to 4) with
  output preserved in input order
- One bad username does not kill the batch; per-user errors render in
  both text and JSON output, and the process exits non-zero
- Hound failures degrade to a warning instead of dropping org info
- `--output json` for scripting

OWNERS matching:
- Anchors the hound query on "- <user>" and keeps a hit only when the
  YAML list item equals the username exactly, so prose mentions,
  commented-out entries and substring near-misses ("- alice-bot" for
  "alice") are all dropped. Surrounding quotes and trailing inline
  comments are stripped first, so `- "alice"` and "- alice # area expert"
  still match
- Each surviving hit is confirmed against the file's real content via
  `gh api` to exclude emeritus-only entries, toggled by --verify-owners.
  A hit that cannot be fetched is kept with a warning rather than dropped
- Results render as an aligned REPO / PATH / URL table; the URL uses
  /blob/HEAD/ so the link resolves to the default branch

Tests:
- Table-driven coverage of the OWNERS match filter (quoted values, inline
  comments, commented-out entries, substring near-misses), emeritus
  exclusion, and the fail-open path when a hit cannot be verified
- Batch coverage for both output modes: input ordering, mixed
  success/failure runs, and the text renderer's company / orgs fallbacks
- Asserts hound requests carry no Authorization header while GitHub
  requests stay authenticated

Dependencies:
- adds github.com/google/go-github/v88 for the user API
- promotes golang.org/x/sync from indirect v0.2.0 to direct v0.8.0 for
  errgroup
- go directive 1.22.4 -> 1.25.0

(This PR has gotten assistance from AI tools)

### DIFF
--- a/cmd/korg/korg.go
+++ b/cmd/korg/korg.go
@@ -17,10 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/spf13/cobra"
 )
@@ -58,6 +61,27 @@ Note: Removing from teams is currently unsupported.
 	`
 
 	auditHelpText = "Audit GitHub org members"
+
+	userinfoHelpText = `
+Gets k8s org membership, GitHub profile, and OWNERS file references for user(s).
+
+Honors GITHUB_TOKEN / GH_TOKEN for authenticated GitHub API access.
+
+A failed per-user lookup does not omit that user from the output: in text mode
+it renders an ERROR block, and in JSON mode it appears with only "username"
+and "error" populated. Check the exit code to detect any failures.
+
+By default, each OWNERS/OWNERS_ALIASES hit is confirmed against the file's
+actual content (via the gh CLI) to exclude emeritus-only entries; this costs
+one "gh api" call per hit and requires the gh CLI to be installed and
+authenticated. Pass --verify-owners=false to skip this and use hound's raw
+search hits as-is.
+
+	korg userinfo <github username>
+	korg userinfo <github username1> <github username2> <github username3> ...
+	korg userinfo --output json <github username>
+	korg userinfo --verify-owners=false <github username>
+	`
 )
 
 type Options struct {
@@ -222,7 +246,32 @@ func main() {
 	rootCmd.AddCommand(removeCmd)
 	rootCmd.AddCommand(auditCmd)
 
-	if err := rootCmd.Execute(); err != nil {
+	var (
+		outputFormat string
+		verifyOwners bool
+	)
+	userInfoCmd := &cobra.Command{
+		Use:   "userinfo",
+		Short: "Get information about user(s)",
+		Long:  userinfoHelpText,
+		Args:  cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			switch outputFormat {
+			case "text", "json":
+			default:
+				return fmt.Errorf("invalid --output %q (want: text, json)", outputFormat)
+			}
+			return runUserinfo(cmd.Context(), o.RepoRoot, args, outputFormat == "json", verifyOwners, cmd.OutOrStdout())
+		},
+	}
+	userInfoCmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "output format: text|json")
+	userInfoCmd.Flags().BoolVar(&verifyOwners, "verify-owners", true, "confirm each OWNERS/OWNERS_ALIASES hit via the gh CLI and exclude emeritus-only entries; requires gh to be installed and authenticated")
+	rootCmd.AddCommand(userInfoCmd)
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
 		os.Exit(1)
 	}
 }
--- a/cmd/korg/userinfo.go
+++ b/cmd/korg/userinfo.go
@@ -1,0 +1,463 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/google/go-github/v88/github"
+	houndclient "github.com/hound-search/hound/client"
+	houndindex "github.com/hound-search/hound/index"
+	"golang.org/x/sync/errgroup"
+	"sigs.k8s.io/prow/pkg/config/org"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	houndSearchURL      = "https://cs.k8s.io/api/v1/search"
+	defaultHTTPTimeout  = 30 * time.Second
+	userinfoConcurrency = 4
+
+	// unknownLookupError is reported when a lookup yields neither a result nor an
+	// error. findUserDetails never returns that combination today; both output modes
+	// carry the fallback so that if the invariant ever breaks, the user is still
+	// reported instead of silently disappearing.
+	unknownLookupError = "unknown error"
+)
+
+// Test-only overrides. Empty means use defaults / api.github.com.
+var (
+	houndSearchURLOverride string
+	ghBaseURLOverride      string
+)
+
+func houndURL() string {
+	if houndSearchURLOverride != "" {
+		return houndSearchURLOverride
+	}
+	return houndSearchURL
+}
+
+type OrgMembership struct {
+	Org  string `json:"org"`
+	Role string `json:"role"` // "member" or "admin"
+}
+
+type OwnerFile struct {
+	Repo string `json:"repo"`
+	Path string `json:"path"`
+	URL  string `json:"url"`
+}
+
+type UserDetails struct {
+	Username   string          `json:"username"`
+	Company    string          `json:"company,omitempty"`
+	Orgs       []OrgMembership `json:"orgs"`
+	OwnerFiles []OwnerFile     `json:"owner_files,omitempty"`
+	Warnings   []string        `json:"warnings,omitempty"`
+	Error      string          `json:"error,omitempty"`
+}
+
+// findUserDetails gathers GitHub profile, k8s org membership, and OWNERS file references for a username.
+// GitHub or k/org config failures are fatal (returned as error). Hound failures are non-fatal warnings.
+// When verifyOwners is true, each OWNERS hit is additionally confirmed against the file's actual
+// content (via the gh CLI) to exclude emeritus-only entries; this costs one gh api call per hit.
+func findUserDetails(ctx context.Context, gh *github.Client, hc *http.Client, configs map[string]*org.Config, username string, verifyOwners bool) (*UserDetails, error) {
+	info := &UserDetails{Username: username}
+
+	u, _, err := gh.Users.Get(ctx, username)
+	if err != nil {
+		return nil, fmt.Errorf("github user %q: %w", username, err)
+	}
+	if u.Company != nil {
+		info.Company = *u.Company
+	}
+
+	info.Orgs = findOrgMembership(configs, username)
+
+	hits, err := searchOwnerFiles(ctx, hc, username)
+	switch {
+	case err != nil:
+		info.Warnings = append(info.Warnings, fmt.Sprintf("OWNERS lookup failed: %v", err))
+	case !verifyOwners:
+		info.OwnerFiles = hits
+	default:
+		info.OwnerFiles = make([]OwnerFile, 0, len(hits))
+		for _, hit := range hits {
+			content, err := fetchOwnerFileContent(ctx, hit.Repo, hit.Path)
+			if err != nil {
+				// Fail open: we can't confirm the hit, but hound already found it, so keep it.
+				info.Warnings = append(info.Warnings, fmt.Sprintf("could not verify %s/%s: %v", hit.Repo, hit.Path, err))
+				info.OwnerFiles = append(info.OwnerFiles, hit)
+				continue
+			}
+			if isActiveOwner(content, hit.Path, username) {
+				info.OwnerFiles = append(info.OwnerFiles, hit)
+			}
+		}
+	}
+
+	return info, nil
+}
+
+func findOrgMembership(configs map[string]*org.Config, username string) []OrgMembership {
+	out := []OrgMembership{}
+	orgNames := make([]string, 0, len(configs))
+	for name := range configs {
+		orgNames = append(orgNames, name)
+	}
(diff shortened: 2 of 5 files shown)


cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
Priyankasaggu11929:
```suggestion
Copyright The Kubernetes Authors.
```

per https://github.com/kubernetes/kubernetes/commit/c247514f47f4331b82b9b6ca374b9a7a85690b57

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	return out, nil
+}
+
+func ownerFileURL(repo, path string) string {
+	return fmt.Sprintf("https://github.com/%s/blob/HEAD/%s", repo, path)
+}
+
+func searchOwnerFiles(ctx context.Context, hc *http.Client, username string) ([]OwnerFile, error) {
+	q := url.Values{}
+	q.Set("q", username)
```
Priyankasaggu11929: should we do a search for `"- " + username` instead of just `username`?

since OWNERS entries are always YAML list items

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	q.Set("repos", "*")
+	q.Set("rng", ":20")
+	q.Set("files", "OWNERS")
```
Priyankasaggu11929: let's include `OWNERS_ALIASES` as well

since that defines named groups (e.g. sig-foo-approvers) which are referenced in OWNERS files

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	for repo, matches := range r.Results {
+		if matches == nil {
+			continue
+		}
+		for _, fm := range matches.Matches {
+			out = append(out, OwnerFile{
+				Repo: repo,
+				Path: fm.Filename,
+				URL:  ownerFileURL(repo, fm.Filename),
+			})
```
Priyankasaggu11929, line 221: can we add a check here to skip lines starting with `#` in the result before appending them to `out`?

because commented out usernames also will show up as valid hit right now

Also -

(since we already have the repo + file path from hound) 
could we fetch the actual OWNERS file from GitHub and parse which section the username falls under? because someone in `emeritus_approvers/emeritus_reviewers` isn't valid


it will add 1-2 extra API calls maybe but since its supposed to be used locally on a handful of users at a time, it think that should be ok

cmd/korg/userinfo_test.go
```diff
@@ -0,0 +1,286 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
Priyankasaggu11929:
```suggestion
Copyright The Kubernetes Authors.
```

per https://github.com/kubernetes/kubernetes/commit/c247514f47f4331b82b9b6ca374b9a7a85690b57

Priyankasaggu11929 (changes requested):
Thank you, @palnabarun!

left a few inline comments.

go.mod
```diff
@@ -1,89 +1,97 @@
...
-	github.com/sirupsen/logrus v1.9.0
-	github.com/spf13/cobra v1.7.0
-	k8s.io/apimachinery v0.27.4
-	sigs.k8s.io/prow v0.0.0-20240418142548-4c9d8ca1213d
-	sigs.k8s.io/yaml v1.5.0
+	github.com/sirupsen/logrus v1.9.3
+	github.com/spf13/cobra v1.10.2
+	golang.org/x/sync v0.20.0
+	k8s.io/apimachinery v0.33.11
+	k8s.io/test-infra v0.0.0-20191222193732-de81526abe72
```
cblecker: I'm not sure what happened here, but you're pulling in a whole bunch of new/different stuff that seems unrelated, for example, the test-infra repo?

go.mod
```diff
@@ -1,89 +1,97 @@
...
+	github.com/spf13/pflag v1.0.9 // indirect
+	github.com/xanzy/ssh-agent v0.3.3 // indirect
+	golang.org/x/crypto v0.50.0 // indirect
+	golang.org/x/net v0.53.0 // indirect
+	golang.org/x/oauth2 v0.34.0 // indirect
+	golang.org/x/sys v0.43.0 // indirect
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+)
+
+// Pin all k8s.io staging repositories to kubernetes-1.15.3.
```
cblecker: This isn't accurate?

cmd/korg/korg.go
```diff
@@ -222,7 +235,29 @@ func main() {
...
+		},
+	}
+	userInfoCmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "output format: text|json")
+	rootCmd.AddCommand(userInfoCmd)
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
+		fmt.Fprintln(os.Stderr, "error:", err)
```
cblecker: Cobra already prints command errors to stderr by default, so this line will double-print every error. Either drop it to preserve the previous behavior, or set `rootCmd.SilenceErrors = true` before `ExecuteContext` if you want to own the formatting yourself.

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	for _, warn := range u.Warnings {
+		fmt.Fprintln(w, "Warning:", warn)
+	}
+}
+
+// newGitHubClient returns a GitHub client authenticated via GITHUB_TOKEN or GH_TOKEN if set.
+func newGitHubClient(ctx context.Context, httpClient *http.Client) (*github.Client, error) {
+	token := os.Getenv("GITHUB_TOKEN")
```
cblecker, line 373: `ctx` isn't used anywhere in this function — it only reads env vars and constructs a client. Either drop it from the signature or use `_` if you're planning to need it later.

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	results := make([]*UserDetails, len(usernames))
+	errs := make([]error, len(usernames))
+
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(userinfoConcurrency)
+	for i, name := range usernames {
+		i, name := i, name
+		g.Go(func() error {
```
cblecker, line 416: This loop-variable capture was needed before Go 1.22, but since the module is already at 1.22+ (and this PR bumps it further), loop variables are per-iteration by default. This line can be removed.

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+			if err != nil {
+				errs[i] = err
+				return nil
+			}
+			results[i] = info
+			return nil
+		})
+	}
+	_ = g.Wait()
+
```
cblecker, line 429: Since the goroutines always return `nil` today, discarding the `Wait()` return is technically safe — but it's a maintenance trap. If someone later changes a goroutine to return a real error, it'll vanish silently. A small guard makes the invariant explicit:

```go
if err := g.Wait(); err != nil {
    return fmt.Errorf("unexpected errgroup error: %w", err)
}
```

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,306 @@
...
+	if outputJSON {
+		clean := make([]*UserDetails, 0, len(results))
```
cblecker: When a user lookup fails in JSON mode, that user is silently dropped from the output array. A script parsing stdout will get fewer results than expected, with no structured way to tell which lookups failed or why. Text mode handles this well (it renders an `ERROR:` block inline), but JSON consumers have to correlate via exit code alone.

Worth either including an error entry in the array (e.g. `{"username":"ghost","error":"..."}`) or at least documenting in the help text that JSON mode omits failures and callers should check the exit code.

cblecker (commented):
Nice addition — this fills a genuine gap and the overall structure is solid. The errgroup concurrency pattern is correct, Hound failures degrade gracefully, and it's good to see tests land alongside the new code. A few things worth addressing before this merges, mostly around the JSON output dropping failed users and a double-print in `main()`. The go.mod changes are covered separately.

cmd/korg/userinfo.go
palnabarun: I will fix this in a separate PR. There are more changes to do to support this.

cmd/korg/userinfo_test.go
palnabarun: I will fix this in a separate PR. There are more changes to do to support this.

palnabarun:
@Priyankasaggu11929 @cblecker PTAL again. Thank you. 🙏🏽

cmd/korg/userinfo.go
palnabarun: https://github.com/kubernetes/org/pull/6485

cmd/korg/userinfo_test.go
palnabarun: https://github.com/kubernetes/org/pull/6485

cmd/korg/userinfo_test.go
cblecker, line 1: The batch path is the most complex control flow here but only gets tested with single-user inputs. A mixed-batch test (e.g. `["alice", "ghost", "bob"]` where ghost 404s) would exercise both branches of the JSON assembly loop in a single call and catch ordering / nil-dereference regressions.

cblecker, line 1: Text mode is the default output but has no test coverage -- all `runUserinfo` tests pass `outputJSON=true`. A `TestRunUserinfoText` that checks for the `=== username` header, `Company:` line, `Orgs:` section, and tabwriter table would guard against formatting regressions in `renderText`.

cmd/korg/userinfo.go
cblecker: +1 on the error-entry design. One thing worth hardening: the `else` branch on line 409 calls `errs[i].Error()` without checking that `errs[i]` is non-nil. Today `findUserDetails` always returns either `(non-nil, nil)` or `(nil, error)`, so the invariant holds -- but if it's ever refactored to return `(nil, nil)`, this panics. A nil guard (`else if errs[i] != nil { ... } else { ... "unknown error" }`) would make this robust for free.

cmd/korg/userinfo.go
cblecker: The `"- "` prefix was a good fix. One edge: since hound does substring matching, `"- alice"` also hits lines containing `"- alicexyz"` or `"- alice-bot"`. With `--verify-owners=true` (the default) this is filtered out by the YAML parse, but with `--verify-owners=false` those false positives pass through. Could tighten `hasUncommentedMatch` to also check the trimmed line after `"- "` matches the username exactly -- or just document this as a known limitation of the unverified mode.

palnabarun:
@cblecker addressed all four: nil-guard on the JSON error entry, tightened the "- <user>" match to require an exact username (not just substring) after the dash, and added tests for the mixed-batch case and the default text-mode renderer. PTAL 🙏🏽

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,441 @@
...
+
+// runUserinfo fetches info for every username concurrently and writes ordered output to w.
+// Returns a joined error of all per-user failures; users without errors still render.
+func runUserinfo(ctx context.Context, repoRoot string, usernames []string, outputJSON, verifyOwners bool, w io.Writer) error {
+	configs, err := loadOrgConfigs(repoRoot, validOrgs)
+	if err != nil {
+		return err
+	}
+
+	hc := &http.Client{Timeout: defaultHTTPTimeout}
```
cblecker: The `hc` client is shared between `newGitHubClient` and `searchOwnerFiles`. The go-github library mutates the provided `http.Client`'s Transport in-place to inject `Authorization: Bearer <token>` headers. After `newGitHubClient(hc)` returns, every subsequent `hc.Do(req)` — including hound requests to `cs.k8s.io` — sends the GitHub token.

Fix: use separate `http.Client` instances for GitHub API and hound requests.

cmd/korg/userinfo.go
```diff
@@ -0,0 +1,441 @@
...
+	hits, err := searchOwnerFiles(ctx, hc, username)
+	switch {
+	case err != nil:
+		info.Warnings = append(info.Warnings, fmt.Sprintf("OWNERS lookup failed: %v", err))
+	case !verifyOwners:
+		info.OwnerFiles = hits
+	default:
+		info.OwnerFiles = make([]OwnerFile, 0, len(hits))
+		for _, hit := range hits {
+			content, err := fetchOwnerFileContent(ctx, hit.Repo, hit.Path)
```
cblecker, line 115: The fail-open behavior here (keep the hit + add a warning when verification fails) is a deliberate design choice, but no test exercises it. If someone refactored this to drop unverifiable hits, no test would break.

Consider adding a subtest where `stubGHCLI` returns an error for one of two hits, then assert both hits are preserved and a warning is emitted.

cblecker (changes requested):

cmd/korg/userinfo.go
cblecker: Following up on the matching improvements — `hasRealMatch` strips quotes via `strings.Trim(rest, "\"'")`, but no test exercises quoted YAML values like `- "alice"` or `- 'alice'`. A dedicated `TestHasRealMatch` table-driven test would cover this and provide a clearer safety net for future changes to the filtering logic.

cmd/korg/userinfo_test.go
cblecker: `TestRunUserinfoText` covers the happy path now. A few edge branches in text mode are still untested: the ERROR block (lines 433-435), the `Company: **Not Found**` fallback for empty company, and the `(none)` display for empty orgs. A mixed-batch text test (similar to `TestRunUserinfoJSONMixedBatch` but with `outputJSON=false`) would close the gap.

cmd/korg/userinfo.go
cblecker: One more edge case: YAML inline comments like `- alice # area expert` would fail the `EqualFold` check because the comment text is included in the comparison. Stripping inline comments before the equality check would handle this: `if idx := strings.Index(rest, "#"); idx >= 0 { rest = strings.TrimSpace(rest[:idx]) }`. Unlikely in practice since OWNERS files rarely use inline comments, but it's a silent false negative if it does.

cmd/korg/userinfo.go
cblecker: The JSON mode now has a defensive `default: "unknown error"` handler for the (nil, nil) case. Text mode (lines 429-437) uses two independent if-checks with no fallback, so a user with both values nil would be silently omitted. Adding a parallel fallback in text mode would maintain parity.

cmd/korg/userinfo_test.go
```diff
@@ -0,0 +1,824 @@
...
+
+func TestFindUserDetails(t *testing.T) {
+	stubGHCLI(t, func(ctx context.Context, args ...string) ([]byte, error) {
+		if len(args) == 4 && args[0] == "api" && args[1] == "repos/k/k/contents/OWNERS" {
+			return []byte("approvers:\n- alice\n"), nil
+		}
+		return nil, fmt.Errorf("unexpected gh invocation: %v", args)
+	})
+
+	ghHits := 0
```
cblecker, line 237: nit: `ghHits` is incremented but never read or asserted on — it's dead code. Either remove it or add an assertion like `if ghHits == 0 { t.Error("expected at least one GitHub API call") }` to make the intent explicit.

PR #6399, opened 2026-05-24
Add v1.37 release branch management team members

aibarbetta:
I'm assuming branch management members just need to be here to get access to the release google docs. For other branch management tasks we'll update [this group](https://github.com/kubernetes/k8s.io/blob/main/groups/sig-release/groups.yaml#L33)


dipesh-rawat (commented):
/lgtm

fsmunoz (approved):

PR #6414, opened 2026-05-28
Update PRR shadows for 1.37 cycle

soltysh:
Following May 27th PRR call (https://docs.google.com/document/d/10QkXwiZfBL7wBKHHFslooXvGDKHA75jkkgiIXuKFryM/edit?tab=t.0) this updates the PRR shadows for 1.37 cycle. 

/assign @johnbelamaric

### DIFF
--- a/config/kubernetes/sig-architecture/teams.yaml
+++ b/config/kubernetes/sig-architecture/teams.yaml
@@ -94,13 +94,20 @@ teams:
       prod-readiness-reviewers:
         description: Production Readiness Reviewers
         members:
-        - aramase       # shadow
-        - deads2k       # approver
-        - johnbelamaric # approver
-        - jpbetz        # approver
-        - kannon92      # approver
-        - sohankunkerkar # shadow
-        - soltysh       # approver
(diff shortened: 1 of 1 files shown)


config/kubernetes/sig-architecture/teams.yaml
```diff
@@ -94,13 +94,19 @@ teams:
       prod-readiness-reviewers:
         description: Production Readiness Reviewers
```
Priyankasaggu11929, line 95: if the gh team is used for co-ordination, then we can do this

```suggestion
        description: Production Readiness Reviewers
        maintainers:
        - palnabarun #shadow
        - Priyankasaggu11929 #shadow
```

dims:
/approve

in case you want to tweak things and then lgtm

config/kubernetes/sig-architecture/teams.yaml
soltysh: It's not needed, since I know you already have access to that board, so it's not necessary.

PR #6536, opened 2026-08-07
Add damikag to kubernetes-sigs

damikag:
I'm a member of Kubernetes org in https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler. We're migrating to a separate repository in kubernetes-sigs org (https://github.com/kubernetes-sigs/cluster-autoscaler), so I need to propagate my membership.

### DIFF
--- a/config/kubernetes-sigs/org.yaml
+++ b/config/kubernetes-sigs/org.yaml
@@ -233,6 +233,7 @@ members:
 - damdo
 - damemi
 - DamianSawicki
+- damikag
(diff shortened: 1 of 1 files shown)


Priyankasaggu11929:
existing member - https://github.com/kubernetes/org/blob/8e36256227612c0f97a803cb1000a26857b953f6/config/kubernetes/org.yaml#L266

/lgtm
/approve

Thanks!

PR #6621, opened 2026-09-08
Sync SIG Scheduling groups

macsko:
Sync new reviewers and approvers in the SIG Scheduling teams


config/kubernetes/sig-scheduling/teams.yaml
dom4ha, line 2: does that group impose any permissions or is it just used in mentioning the whole team `@kubernetes/sig-scheduling-misc`?

macsko: AFAIK, it's used only for mentioning the team

PR #6632, opened 2026-09-10
Add spowelljr to minikube-preloads teams

spowelljr:


neolit123 (commented):
please ask @medyagh to lgtm this PR.

yujuhong:
@neolit123 quick context on why we need the admin permissions for Steven (@spowelljr ): 

The current owner is on extended leave. We're working to keep Minikube afloat and and also transition out the Google-owned release pipeline. Steven has deep context here as a past root approver and current component admin, and he just cut the v1.39 release for minikube. Granting him access will unblock the ongoing minikube infra work.

neolit123:
that's fine @yujuhong . i just needed someone trusted to confirm. thanks for that! might be worth adding at least one more person to this team.

/lgtm
/approve

PR #6653, opened 2026-09-17
chore: add serngawy to CAPA release group

richardcase:
This updates the release group for CAPA to remove damdo and add serngawy. It also changes the permissions on the group to remove write.

/cc @damdo


damdo (commented):
/approve
/lgtm

PR #6657, opened 2026-09-19
Move myself to emeritus

rexagod:
Going forward, as discussed internally within the SIG, I will still be around every now and then to help out efforts that are currently in an incubation phase, and were signed-off on during my tenure as a chair.

I've followed earlier practices and only removed my handle from areas where previous chairs did that as well. Please let me know if I missed something.

cc @kubernetes/sig-instrumentation-leads


config/kubernetes/sig-instrumentation/teams.yaml
```diff
@@ -5,7 +5,6 @@ teams:
     - dashpole
     - dgrisonnet
     - pohly
-    - rexagod
     - richabanker
```
dashpole: Did you also want to remove yourself from the SIG-level admin teams in `config/kubernetes-sigs/sig-instrumentation/teams.yaml` (like `instrumentation-admins`, `instrumentation-tools-admins`, and `instrumentation-tools-maintainers`), matching what `ehashman` did in `f8fb7cc6`?

